PR #9, opened 2015-03-10
Add a --disable-plugins option to ./configure (Fixes #8)

### LINKED ISSUE #8: Add a --disable-plugins option?
Is this possible?

I'm not sure (*from reading the sources*) that any of the plugins are gonig to work for our system design. Just tested a compile against HEAD:

```#!bash
plugins/bootmisc.c:32:2: error: #error This plugin uses getuser(), which relies on dynamically linked GLIBC. Cannot be built statically!
 #error This plugin uses getuser(), which relies on dynamically linked GLIBC.  Cannot be built statically!
  ^
common.mk:18: recipe for target 'plugins/bootmisc.o' failed
make: *** [plugins/bootmisc.o] Error 1
=======> ERROR: Building '/home/prologic/vallinux/base/packages/finit/finit#HEAD-1.pkg.tar.gz' failed.
```

therealprologic:


therealprologic:
Cool :)

PR #12, opened 2015-10-16
Unblock blocked signals after forking off a child

wkz:
After forking off a new child, all signals that are blocked by finit
should be unblocked before exec-ing the final binary. Also, finit now
blocks SIGHUP, so add it to the set of signals that should be
unblocked.


troglobit:
Thx!

PR #16, opened 2015-12-01
Fix to get finit to work with BusyBox getty

villains:
Hi Joaquim,
I was having trouble getting finit to show a login prompt on a serial console on my BusyBox system. Taking a look, I noticed that the command line constructed for the BusyBox getty program was the problem. The fix was pretty simple.
Thanks for your work on finit.
All the best,
Markus

### DIFF
--- a/tty.c
+++ b/tty.c
@@ -130,7 +130,7 @@ tty_node_t *tty_find_by_pid(pid_t pid)
 void tty_start(finit_tty_t *tty)
 {
 	int i = 0, is_console = 0;
-	char *cmd, *arg;
+	char *cmd, *arg, *name;
 	char  line[FILENAME_MAX];
 	char *args[42];
 
@@ -144,7 +144,9 @@ void tty_start(finit_tty_t *tty)
 #if   defined(GETTY_AGETTY)
 		snprintf(line, sizeof(line), "%s %s %d %s", GETTY, tty->name, tty->baud, tty->term ?: "");
 #elif defined(GETTY_BUSYBOX)
-		snprintf(line, sizeof(line), "%s %d %s %s", GETTY, tty->baud, tty->name, tty->term ?: "");
+		/* BusyBox getty won't work if the TTY device name is prefixed with /dev path */
+		name = tty->name + strlen(_PATH_DEV);
+		snprintf(line, sizeof(line), "%s %d %s %s", GETTY, tty->baud, name, tty->term ?: "");
 #else
 		strlcpy(line, FALLBACK_SHELL, sizeof(line));
 		is_console = 1;


villains:
I'm running BusyBox 1.23.1.
You're right, based on the code, it certainly looks like it should accept devices with and without /dev prefixes. It would appear that for some reason, the code is not finding '/' as the first char in G.tty_name.

villains:
I'm comparing my BusyBox configuration with your TroglOS config-1.23 file to see if I can spot anything.

troglobit:
Good idea diffing the configs!

It seems the only thing that can spoil the party is `NOT_LONE_DASH()`

    #define NOT_LONE_DASH(s) ((s)[0] != '-' || (s)[1])

But that shouldn't apply here.  Could it be some nasty unicode copy-paste thing?

villains:
Good thought. I checked, but they are real front slashes, not some UTF-8 front slash look-alike. :)

villains:
I reverted back to commit 7027cbc, made the change to finit.conf that you suggested, and booted up my board. With "finit_debug" on the kernel cmd line, here is what I got:

Please press Enter to activate this console.
finit:sig.c:sigchld_cb() - Collected child 695
finit:tty.c:tty_start() - Starting console: /sbin/getty on /dev/ttyO2

Please press Enter to activate this console.


Notes:
- system seems to boot up fine
- finit seems to be calling BusyBox getty correctly
- BusyBox getty seems to be crapping out after several seconds, and returning control back to finit

villains:
Right now this is looking more like a problem in BusyBox, not finit.

troglobit:
Very weird indeed.  You could try putting in some more `_d()` debug lines, prior to my chopping up `line` perhaps.  Just to check ... but I agree with you, that looks like a problem with BusyBox ...

Could it be some kernel tty IOCTLs that getty use to probe? (I think it does this ...)

troglobit:
Btw. one of the reasons for creating TroglOS was to have a testing environment for as close to vanilla kernel, busybox, etc. as possible, so I could test my own applications.  But some also use it for testing new applications before targeting a real embedded system.  

So maybe test your kernel config + busybox config + finit.conf on TroglOS to see if you can reproduce the problem?

villains:
Another weird thing. I set up finit using the following configuration:

>user root
>host dingle
>runlevel 2
>service [S12345] /sbin/syslogd -n -b 20 -s 1024 -D -O /var/syslog    -- System log daemon
>service [S12345] /sbin/klogd -n                              -- Kernel log daemon
>service :1 [2345] <!IFUP:eth0,GW:UP> /sbin/dropbear -R -F -p 22 -- SSH daemon
>tty [12345] ttyO2 115200 vt100
>console ttyO2

(Some commented rows removed for clarity.)

The system hung after finit started up.
I could see where finit registered ttyO2, but there was no message saying finit was starting getty. In previous attempts, I could always see where finit was starting getty.

The attached file contains the full output from finit during the boot:
[20151201-finit-startup.txt](https://github.com/troglobit/finit/files/49048/20151201-finit-startup.txt)

Any thoughts?

P.S. This was with finit built from commit 7027cbc (no changes by me).

villains:
I'll look at trying TroglOS to see if this can be reproduced there.

villains:
For comparison, here is the finit debug output when using "/dev/ttyO2" instead of "ttyO2".
[20151201-finit-startup-2.txt](https://github.com/troglobit/finit/files/49069/20151201-finit-startup-2.txt)

Here finit does seem to start getty.
However, the previous problem is back, where getty stops running after several seconds without displaying a login prompt, and finit displays the message "Please press Enter..." again.

troglobit:
Hmm, that hang I've actually heard from some people at work .. will look into that during the day tomorrow (CET), thank you for mentioning it!  Means it could actually be Finit and none of the local adaptations we've done at work (lots of plugins and such).

If any service repeatedly fails to start (including tty's) Finit halts them and you have to manually do "initctl restart NAME" to clear the restart counter.  

Looking forward to if you can reproduce your problem with TroglOS, thanks!

villains:
Joaquim, I found the issue.
Somehow in my rootfs image recipe, util-linux-agetty was in the list of installed packages. Basically I was having two getty's installed on my system, and unluckily the non-BusyBox one was getting invoked by finit. With the package removed, finit starts up and the login prompt appears exactly as it should.

I guess the lesson is to look carefully at what is getting into your rootfs. :D
I will close my finit fork and remove the pull request.
Thank you for your kind assistance, and sorry for the false alarm.

troglobit:
Awesome job finding it! :+1: 

At first I considered asking you about that, but I found it too far fetched, I'd rather consider Finit to be buggy than question any of my user's ;-)

Good luck with your Gumstix, and don't hesitate to check back with any updates or questions again!

Cheers!
 /Joachim :smiley:

villains:
Yes... it was quite a facepalm moment for me, Joaquim 😱 😏
I did have a couple of ideas for features, which I will ask about in a separate thread.
All the best,
Markus

troglobit:
No worries mate! :smiley: 

Looking forward to them!

troglobit:
Just FYI, I just pushed two fixes to the `initctl reload` logic which should fix the last of
the hang problems people have reported.

PR #26, opened 2016-04-13
Some fixes, and add logging

patstew:
I've made some fixes so it works with musl libc, fixed a problem with 'run' and added logging of service's stdout.
I haven't actually built or checked this specific version, because the autogen stuff is complaining about libuev now, but before I rebased it was fine.


patstew:
The commit message of 	3de9e4e is a bit OTT, it probably won't hang, I may have been blaming it for an issue with my script. I think there is a race though.

troglobit:
Looks great, good sense of style, neatly packaged and not everything in one big lump.
Awesome job! :smiley: :+1:

PR #62, opened 2017-01-11
Fix fluxstate

mattiaswal:
There are some problems with applications stay in flux state/never become flux state.

* If an appliciation depend on an application run later in the service list it will be kept in flux
* When running on X86, finit sometimes complete its operation on below 1 s, the timestamp used to detect if any service is dirty is only using seconds.

### DIFF
--- a/plugins/pidfile.c
+++ b/plugins/pidfile.c
@@ -30,6 +30,7 @@
 #include "../cond.h"
 #include "../helpers.h"
 #include "../plugin.h"
+#include "../service.h"
 
 struct context {
 	int fd;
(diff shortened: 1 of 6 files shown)


plugins/pidfile.c
```diff
@@ -88,16 +89,28 @@ static void pidfile_callback(void *UNUSED(arg), int fd, int UNUSED(events))
...
 	(void)(_null);
+	int restart = 0;
+
+	do {
+		restart = 0;
 
-	for (svc = svc_iterator(1); svc; svc = svc_iterator(0)) {
-		if (svc->state == SVC_RUNNING_STATE && !svc_is_changed(svc) && !svc_is_starting(svc)) {
+		for (svc = svc_iterator(1); svc; svc = svc_iterator(0)) {
 			snprintf(name, MAX_ARG_LEN, "svc%s", svc->cmd);
```
reverendhomer: broken coding style from here

plugins/pidfile.c
```diff
@@ -88,16 +89,28 @@ static void pidfile_callback(void *UNUSED(arg), int fd, int UNUSED(events))
...
-			cond_set_path(cond_path(name), COND_ON);
+			if (svc->state == SVC_RUNNING_STATE &&
+			    !svc_is_changed(svc) &&
```
reverendhomer: //sorry, ignore first comment

coding style broken *here* :)

troglobit: Yup.  I'll merge anyway and fix later, this is an important fix that @mattiaswal has worked on for quite some time.

troglobit: On second thought, that is actually correct coding style.  The Emacs linux-mode usually does the right thing :)

PR #63, opened 2017-01-18
initctl: Add 'initctl cond dump' to show all conditions in the system

mattiaswal:
Remove verbose of initctl cond, and replace it with 'initctl cond dump'
Show all conditions if they are in flux/on even if there are no other
services depending on this condition.

Signed-off-by: Mattias Walström <lazzer@gmail.com>


troglobit:
Looks great, thanks! 👍

PR #70, opened 2017-03-20
initctl -> do_log() need these as runtime depends

abucodonosor:
I'm not sure how this is handled by BB however we need to check for needed progs
and at least warn when not found.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -32,6 +32,20 @@ AC_TYPE_UINT8_T
 AC_TYPE_UINT16_T
 AC_TYPE_UINT32_T
 
+# runtime progs for src/initctl.c:L144
+# should be this AC_MSG_ERROR ?
+AC_CHECK_PROG([grep],[grep],[yes],[no])
+AM_CONDITIONAL([FOUND_GREP], [test "x$grep" = xyes])
+AM_COND_IF([FOUND_GREP],,[AC_MSG_WARN([initclt: required runtime program 'grep' not found.])])
+
+AC_CHECK_PROG([cat],[cat],[yes],[no])
+AM_CONDITIONAL([FOUND_CAT], [test "x$cat" = xyes])
+AM_COND_IF([FOUND_CAT],,[AC_MSG_WARN([initctl: required runtime program 'cat' not found.])])
+
+AC_CHECK_PROG([tail],[tail],[yes],[no])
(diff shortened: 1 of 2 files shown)


troglobit:
This cannot be configure checks, would not work when cross-compiling. 

Wouldn't it be easier if grep/cat/tail was in `$PATH` so we don't have to provide a fully qualified path to them?  

Unless you disagree, and provide a good argument as to why, I'll decline this pull request.

abucodonosor:
Hmm they should work since is just a warning when not found..

Yes they should be in $PATH .. however peoples want initramfs and need to know what tools
need be copyed inside so everything is working.

Right now I know I'll need these while looked at the code.

However it can be documented..

The patches --with-foo will have something like

> AC_ARG_WITH(mount, AS_HELP_STRING([--with-mount=/path/to/mount], [path to mount binary]))
if test "x${with_mount}" = x; then
  AC_PATH_PROG(MOUNT_PATH, mount, [], $PATH:/sbin:/usr/sbin:/bin)
else
  MOUNT_PATH="$with_mount"
fi
AC_DEFINE_UNQUOTED(MOUNT_PATH, "$MOUNT_PATH", [Define to path of mount binary])
AC_SUBST(MOUNT_PATH)

so we can use 

run("MOUNT_PATH  ...");

If you don't like this just tell me how you want to have all these.

troglobit:
`AC_ARG_WITH` is better than `AC_CHECK_PROG`, but the default/fall-back shouldn't be to search for them  in the host's PATH.  They should default to the current hard-coded, and common (Alpine/Debian/BusyBox) default locations for these binaries.

However, what I meant previously, referring to `$PATH`, was that finit itself should probably use `run("mount")` et. al. without an absolute path.  In `finit.c` we set up `$PATH` using `setenv("PATH", _PATH_STDPATH, 1);` ... that should possibly be called earlier so we don't have to hard code absolute paths to these standard tools.

If a distribution has other standard paths than what is defined in `paths.h`, then they have a discrepancy that needs to be fixed there, not in Finit.

abucodonosor:
Ach .. now I get that right..

Ofc that is the better solution and setting up $PATH should solve all this.

troglobit:
It's not always easy to get a point across, in particular in written form. English isn't even my native language.  Glad you understood :)

So, I guess we can close this PR then? Better if you open a new one.

Issue #71, opened 2017-03-23
Allow vendor service file that are not enabled by default


DeX77:
Hi,

referring to the discussion in #69 a way to provide some service definition without automatically enabling it would be nice.

Traditional inits use a file based approach for that: if a symlink to the actual config is found in a runlevel/target specific directory the service will be enabled.

This is "UNIX way" and its easy for a Distro to provide those files, and have a script to either enable the service on install or use a *ctl tool to do it.

The other idea mentioned in #69 (to use a command to create the file) is also possible but not _that_ easy.

Also it might be an idea to have a "override" feature so files in `/etc/finit.d` will override those in `/usr/lib/finit.d` which would be difficult with the command.

troglobit:
Hi!

Yes, this is what I'm considering. Just discussed this with a colleague of mine and what we came up with is something like this:

1. Define a *recommended* directory structure for distributions
2. Document better the possibility to change the location of `/etc/finit.d` to whatever a distribution likes
3. Update man pages
4. Provide examples

Currently I'm leaning towards this layout:

    /etc/finit/
      |-- available.d/*.conf
      `-- enabled.d/dropbear.conf --> ../available.d/dropbear.conf
    /etc/finit.conf

A distribution could then use a script that symlinks (as in the above example), removes symlinks, like Apache `a2ensite`. It could also copy, create from XML/JSON/db, checkout from a version control system, or whatever feels suitable to their domain.

The default location which finit reads, after `/etc/finit.conf`, is possible to change with `configure --with-rcsd=/path/to/dot.d` and `/etc/finit.conf` can be reduced to a few tasks/services needed for bootstrapping the system, but the name of the file itself can also be changed using `configure --with-config=CONFIG`.

DeX77:
Yes, I agree with that aproach and I think its quite intuitive for the usual Unix/Linux user.  

Perhaps, if I find some time I'll try to patch up something, But no promises. :wink:

troglobit:
Just FYI, my primary use-case for Finit is what is encoded in `configure` and `*.h` files.  It is unlikely that I'd be in favor of changing these defaults :)

If you want to contribute anything for this particular issue I'd recommend writing a `DISTRIBUTION.md`, or similar, with instructions for packagers.  And maybe we could update the Debian and Alpine Linux samples.

Cheers

abucodonosor:
@troglobit 

sure defaults ca be to whatever you wish :)

You define whatever it will be default and we can add configure options and some docs for that.

[...]

PR #72, opened 2017-03-28
Some cleanup

abucodonosor:
PATCH 1 + 2 is about the PATh changes discuss from PR #70 , however this solves just the 
run(/bin/foo); -> run('foo"); part.. fexist() function would need  some changes but using
stat() and friends for it would be a bit to much I guess.. 
( maybe we find a solution for that in #71 too )
However I think is not bad to have an fallback for the PATHs in case something breaks 
and having the PATH setup early is not bad too.

PATCH 3 is a Distro thing , however we really should check ifup exists before trying to use it.

Regards

### DIFF
--- a/src/finit.h
+++ b/src/finit.h
@@ -36,6 +36,16 @@
 #include <lite/lite.h>
 #include <uev/uev.h>
 
+/* just in case */
+#ifndef _PATH_STDPATH
+#define _PATH_STDPATH   "/usr/bin:/bin:/usr/sbin:/sbin"
+#endif
+
+#ifndef _PATH_VARRUN
+#define _PATH_VARRUN    "/var/run/"
+#endif
+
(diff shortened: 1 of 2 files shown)


src/finit.h
```diff
@@ -36,6 +36,16 @@
 #include <lite/lite.h>
 #include <uev/uev.h>
 
+/* just in case */
+#ifndef _PATH_STDPATH
+#define _PATH_STDPATH   "/usr/bin:/bin:/usr/sbin:/sbin"
+#endif
+
+#ifndef _PATH_VARRUN
+#define _PATH_VARRUN    "/var/run/"
```
troglobit: This is fine, but I think most distributions are moving towards `/run`, so that they can keep it on a RAM disk.

PR #73, opened 2017-04-25
inetd: remove ntoh on proto

ptruz:
getprotobyname() returns p_proto in host endian, so it should not be
converted with ntoh. (Note: getservbyname() on the other hand does
return s_port in network endian)

Signed-off-by: Petrus Hellgren <petrus.hellgren@westermo.se>


troglobit:
Good catch, and a well written commit message, thanks! :smiley:  :+1:

PR #80, opened 2017-10-01
helpers.h: fix build error: unknown type name ‘va_list’

joerg-krause:
Building finit with a uclibc-ng-toolchain fails with:
```
../src/helpers.h:53:43: error: unknown type name ‘va_list’
 void    printv          (const char *fmt, va_list ap);
```


troglobit:
Thx! \o/

PR #88, opened 2017-12-14
Drop UDP packets from unwanted interfaces

ccb3:
Previously UDP packets were never dropped and if a request came
from a not valid interface inetd got stuck in an inifinite
loop looking at the same packet every time and not dropping it.

Signed-off-by: Lennart Eriksson <lennart.eriksson@westermo.se>


troglobit:
Looks good, thanks! :sunglasses: :+1:

PR #106, opened 2018-03-23
Do not mark inetd connections for deletion when reloading services

jonasj76:


troglobit:
Awesome work, thanks! :+1:

PR #111, opened 2018-05-07
Only restart inetd services when neccessary

jonasj76:
Do not restart a inetd service if the listening interface is changed. Only bring down established connection which are no longer allowed, i.e. do not touch already allowed established connections.

### DIFF
--- a/src/inetd.c
+++ b/src/inetd.c
@@ -122,7 +122,7 @@ static int inetd_stream_peek(int sd, char *ifname)
 	return 0;
 }
 
-static int get_stdin(svc_t *svc)
+static int get_stdin(svc_t *svc, char* iifname)
 {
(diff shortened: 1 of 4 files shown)


troglobit:
Nicely done, big thanks to both you and Pontus for working on this! 😎✌️

troglobit:
It seems the build fails with `--disable-inetd ` https://travis-ci.org/troglobit/finit/jobs/375759298#L1065 

```
finit-service.o: In function `service_step':
/home/travis/build/troglobit/finit/src/service.c:1118: undefined reference to `inetd_stop_children'
collect2: error: ld returned 1 exit status
make[2]: *** [finit] Error 1
```

troglobit:
Awesome, it builds fine now, thanks! :+1:

PR #122, opened 2020-01-30
Switch to using nanosleep in favour of sleep

jackuess:
The remaining sleep time returned by sleep is not accurate enough to be
used as a means to achieve "signal safe"-sleep. nanosleep can be used
in similar fashion instead and is granular enough for our purposes.

This issue was identified during reboot, when Finit is bombarded with
SIGCHLD, aborting the sleep that we do to give daemons a chance to shut
down gracefully. The total delay ended up being less than a second when
in fact two seconds were the intended delay.

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -28,6 +28,7 @@
 #include <string.h>
 #include <stdarg.h>
 #include <stdlib.h>
+#include <time.h>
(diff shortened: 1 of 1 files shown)


troglobit:
Nice and clean, great work on this! :+1:

PR #138, opened 2020-10-20
[RFC] Boot refactor

wkz:
This series refactors the core of Finit with the goal of giving the user more ability to tailor the system to their liking. As a result, it cuts many lines from `main()`, making it easier to get an overview of the boot process. The big stuff:

- Generic cleanup/moving around of dead/misplaced code.

- Remove all automatic mounting of different filesystems at different points in the boot process; just call out to `mount -na` and let the user decide on the exact set of filesystems and options.

- Move `udev`/`mdev` stuff out to a plugin that can be disabled and replaced with whatever the user needs (or doesn't need).

I would really like some feedback on the series as it is now, so that we can integrate these changes before going further.


troglobit:
No feedback, just looks awesome, great work on this! :heart:

PR #141, opened 2020-12-09
Improve build scripting

concatime:
I also fixed the non-POSIX OR operator `test x -o y`.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -12,7 +12,7 @@ endif
 
 # Compat hook
 install-dev:
-	@make -C src install-pkgincludeHEADERS
+	@cd src && $(MAKE) install-pkgincludeHEADERS
 
 # Target to run when building a release
 release: distcheck
--- a/configure.ac
+++ b/configure.ac
@@ -320,6 +320,6 @@ $($LD --version || true)
 ---------------------------------------------
 
 Check the above options and compile with:
- ${MAKE-make}
+ "${MAKE:-make}"
 
 EOF
--- a/contrib/alpine/build.sh
+++ b/contrib/alpine/build.sh
@@ -3,7 +3,7 @@
 echo
 echo "*** Configuring Finit for Alpine Linux"
 
-if [ ! -e configure ]; then
+if ! test -e configure; then
     echo "    The configure script is missing, maybe you're using a version from GIT?"
     echo "    Attempting to run the autogen.sh script, you will need these tools:"
     echo "    autoconf, automake, libtool, pkg-config ..."
@@ -26,7 +26,7 @@ echo "*** Building ..."
 echo
 make
 
-if [ $? -ne 0 ]; then
+if test $? -ne 0; then
     echo
     echo "*** The build failed for some reason"
     echo
--- a/contrib/alpine/finit.d/Makefile.am
+++ b/contrib/alpine/finit.d/Makefile.am
@@ -8,7 +8,7 @@ install-exec-hook:
 	@for file in $(d_LINKS); do				\
 		lnk=$(DESTDIR)$(docsdir)/$$file;		\
 		src=available/$$file;				\
-		if [ -L $$lnk ]; then				\
+		if test -L $$lnk; then				\
 			echo "Skipping $$lnk exists.";		\
 		else						\
 			$(LN_S) $$src $$lnk;			\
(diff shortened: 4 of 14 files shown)


troglobit:
Quite a few changes, without any real motivation than "for consistency" and "improvements".  Could you perhaps tell me a bit more why you want to do this?  What system/shell has been giving problems?

 - My own style is to use `[` rather than `test` and I'm pretty sure that is POSIX compliant and works with all shells out there.  I've tested on Debian `dash` and Busybox `ash`, as available in Alpine Linux
- Why change `make -c foo` to `cd foo; make`?  Finit is solely built for Linux systems, where GNU make is the canonical standard, so portability to other make systems is not an issue.

concatime:
Yeah, sorry, I was in a hurry.

1.

Even tho the second form, `[`, is POSIX compliant, GNU autotools and
gnu-config use solely on the first form, `test`, and not without reason.

> it's common practice in the basic utility scripts to use test

[Source](https://lists.gnu.org/archive/html/automake-patches/2020-08/msg00004.html).

I’ve detailed
[here](https://lists.gnu.org/archive/html/config-patches/2020-08/msg00005.html)
why `test` is supperior. I already made the changes on other repos, like
[CMake](https://gitlab.kitware.com/cmake/cmake/-/merge_requests/4704).

2.

> Finit is solely built for Linux systems, where GNU make is the
> canonical standard

Even tho in practice, GNU Make is the defacto `make` command on most
machines, there is no specification that that guarantees so. For
instance, in my machine, I have `make` symlinked to
[bmake](http://www.crufty.net/help/sjg/bmake.html) (smake is another
alternative). I am able to build libite and libuev with bmake too. And
looking at the [specification of POSIX
make](https://pubs.opengroup.org/onlinepubs/009695399/utilities/make.html),
thre is no `-C` argument. So `cd ... && make` is the portable way. Your
argument is arguably the same for GNU Bash.

> Finit is solely built for Linux systems, where GNU **bash** is the
> canonical standard

But that’s not a reason to not write your scripts in a POSIX compliant
manner. (is /bin/sh bash? is /usr/bin/make gnu make?)


That being said, I just wanted to improve your code base a little bit :)

troglobit:
OK, so basic drive-by pull request.  Thank you, but no thanks.

Improve is a very subjective thing, and since you're likely not going to be here in ten years, I prefer to keep my style on the scripts, since I'm the one who have to maintain them.

And, like I tried to say above.  Portability to other POSIX systems is not a primary concern.  Getting it to build on other Linux systems is, uClinux, musl, and GLIBC-based Linux systems.  Sorry if my bad English was hard to understand, it's not my first language. The scrips are already POSIX compliant, and Finit will never target anything else than Linux systems.  (Because it has heaps of Linux-specific code in it, and really only makes embedded ppl happy.)  If the day comes and we need to support other make implementations than the de facto GNU make, I hope I will be long dead by then.

concatime:
> The scrips are already POSIX compliant

That’s not the case.
`if [ "x$yorn" = "xy" -o "x$yorn" = "xY" ]; then` is not POSIX. And it’s present 6 times. I fixed them in this PR.

> Portability to other POSIX systems is not a primary concern.

What do you mean by "other"? My environment is Linux. It just happens that I don’t use GNU make as my make. Nor coreutils as my tools (busybox and sbase). But it’s still Linux. Still POSIX.

troglobit:
OK, sure. That's not POSIX, but last I tested it worked with Busybox ash, and Debian's dash. Those two make out just about 100% of the POSIX style shells for Linux today. 

You have not stated anywhere what actual problem or bug your pull request addresses. I've successfully built and installed Finit on a heap of embedded systems running various incantations of C libraries, with or without GNU Coreutils or Busybox.  Granted, most of those builds have been cross-compiled, but far from all.  If you want to port Finit to another Linux based system, or variant of make, that's fine. But please state that clearly when presenting the pull request.  And possibly make several pull requests for separate issues.  You're the one who wants a change in my project, so tell me your use-case.  If there are more people out there then there may be some interest from me, but if it's a fringe target with very few users, I see no point in supporting it too.  *I'm only one person in this project and I have to maintain focus, I'm sure you understand.*

I've been around UNIX and Linux for 25+ years, and maintained this project for 10+ of those years, and have never even heard of bmake before. Heard of many others, but not that one.  Googling a bit it seems to be the same make that I know as pmake in NetBSD ... and that is one hairy beast I don't want to see again.

Also, just because I offer an alternative to systemd doesn't mean I'm prepared to bend over backwards to support any weird system based on Linux out there. There are a lot ...

concatime:
See, that’s exactly why `[` is bad. It is not a shell operator! It has nothing to do with `busybox ash` nor `dash`. I explained it [here](https://lists.gnu.org/archive/html/config-patches/2020-08/msg00005.html). `[` is a command from GNU coreutils `'/usr/bin/['`. If you use the test command from [sbase](https://core.suckless.org/sbase/) instead, you will encounter `too many arguments`.
That being said, some shells, to improve speed, come with builtins (`[`, `test`, `echo`, etc.).

PR #142, opened 2021-02-02
Handle SIGPWR like SIGUSR2

jackuess:
This seems to be inline with Busybox init, SysV init and, according
to the comments in code, the intended behaviour of Finit.

The effect is that `SIGPWR` will lead to Finit issuing the syscall `reboot`
with `cmd` parameter set to `RB_POWER_OFF` (rather than `RB_AUTOBOOT`), which
in turn will lead to a halt rather than a reboot.
The effect of the above inside a PID namespace other than the initial PID
namespace is that `wait` inside the parent of Finit will report that Finit was
killed with a `SIGINT` signal rather than `SIGHUP`.

### DIFF
--- a/src/sig.c
+++ b/src/sig.c
@@ -517,11 +517,11 @@ void sig_setup(uev_ctx_t *ctx)
 
 	/* Standard SysV init calls ctrl-alt-delete handler */
 	uev_signal_init(ctx, &sigint_watcher, sigint_cb, NULL, SIGINT);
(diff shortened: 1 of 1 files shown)


troglobit:
Thanks, great catch this, looks good!

winkmichael:
Good find, glad to see this is getting fixed. We run into it from time to time, but never understood it. 

echo 1 > /proc/sys/kernel/sysrq
echo b > /proc/sysrq-trigger

Has been our occasional work around! haha

PR #144, opened 2021-02-11
Fix #136: drop old init client tool

### LINKED ISSUE #136: Drop old init client tool, we have initctl now

jackuess:
The SysV API compatibility layer gives too little value to be worth
the effort of maintaining it. Users are encouraged to use the
`initctl` tool instead.


src/finit.c
```diff
@@ -368,12 +368,12 @@ int main(int argc, char *argv[])
...
-	 * for compatibility but initctl client tool uses socket
-	 */
-	if (getpid() != 1)
-		return client(argc, argv);
+	if (getpid() != 1) {
+		fprintf(stderr, "Running Finit as anything other than PID 1 "
+			"is not supported.\n"
+			"The initctl tool is the new command line API.\n");
+		return 1;
+	}
```
troglobit, line 376: This is the only comment I have, everything else looks just great!  It is untested, but if we drop this code path entirely we should be able to run as a non-pid1 process monitor.  

However, since that particular use-case is untested, and wasn't part of the original scope for #136, I'll merge this as-is.

troglobit (approved):
Awesome work on this, no changes/follow-ups needed!  Really appreciate you taking the time to also update the README and remove the cruft from the configure script <3

PR #146, opened 2021-02-12
Fix #127: communicate run-parts progress/status

### LINKED ISSUE #127: Always show all scripts executed from rc.local or runparts
... or at least make it configurable.  Many embedded systems use this feature to extend,
or even live-patch systems between major upgrades of firmware.  Security wise it is a
good thing to at least show on the console what scripts are executed outside the normal
bootstrap path.

jackuess:
Printing the progress of run-parts script just like for `rc.local`
seems reasonable. Also log a warning if any script exits abnormally.


troglobit:
Looks great, as usual! <3

PR #153, opened 2021-02-17
Logit post SIGTERM, SIGINT etc.

jackuess:
Since `logit` lives in the same process group as the service, it will receive any signal sent to that group. When a service is stopped and the process group receives `SIGTERM` (or whatever is configured) logit will stop it's business and any output from the service past this point is lost.

This PR moves `logit` outside of the service process group which means that it will live until it's `stdin` is closed, which will happens when the service has actually died.

### DIFF
--- a/src/service.c
+++ b/src/service.c
@@ -191,9 +191,6 @@ static int lredirect(svc_t *svc)
 		return -1;
 	}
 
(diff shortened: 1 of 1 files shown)


troglobit:
Elegant!

PR #156, opened 2021-02-24
Use FINIT_RCSD instead of hard coded value

jackuess:
We need to take `$rcsd_path` into account here right? Otherwise please close :)


troglobit:
Yes :)

PR #157, opened 2021-02-26
Added Edit command

andreasegeberg:
Command Edit <CONF>  opens file /etc/finit.d/available/<CONF>.conf in
the editor referenced by Linux environment variable $EDITOR
If $EDITOR is not set it will try to open emacs and if emacs is not
found it will try to open the file in vim.

Signed-off-by: Andreas Egeberg <andreas.egeberg@gmail.com>


troglobit:
Awesome work!  Travis fails, but that's because it needs a refresh of libite, I'll fix that on master :)

PR #159, opened 2021-03-19
Store removal status separate from dirty

jackuess:
Having removal status stored in the dirty status resulted in removal
status being forgotten when a service was marked as dirty, for instance
when a dependency was updated. This side effect of marke_dirty seems a bit
unexpected and instead of adding exceptions to the logic for when marking
a service dirty - let's separate the two things (dirty and removed) from each
other.

Signed-off-by: Jacques de Laval <Jacques.De.Laval@westermo.com>


troglobit:
LGTM, let's try it out! :)

PR #161, opened 2021-03-20
Fix build for some non-x86 archs

yangfl:
Ref:
https://buildd.debian.org/status/fetch.php?pkg=finit&arch=mips64el&ver=3.2%7Erc3-2&stamp=1609910244&raw=0
https://buildd.debian.org/status/fetch.php?pkg=finit&arch=sparc64&ver=3.2%7Erc3-2&stamp=1609869348&raw=0


troglobit:
This was sort of unexpected, had no idea Finit was making it into Debian! :exploding_head: 

Just a heads-up, we're preparing v4.0 with lots of changes (cgroups v2 and incompatible changes to pid conditions in the pidfile plugin).  

Your patch is obviously being accepted, very neat patch btw.  

Cheers
 /Joachim

PR #165, opened 2021-04-01
Do a full restart rather than HUP if command line of a service has changed

jackuess:
By keeping track of when a full restart is needed, Finit can relieve
the user of having to track modifications to service configuration. This
also makes initctl reload easier to explain and reason about as the effect
of a reload should now be that any configuration updates to services will
be fully "commited" by initctl reload.


troglobit (approved):
Looks great, just like we talked about :+1:

PR #174, opened 2021-05-20
logging: Disregard logging facility when comparing log levels

jackuess:
If we don't extract the level (disregarding facility), statements such as `logit(LOG_CONSOLE | LOG_NOTICE, ...)` would never reach the kernel log.


troglobit:
Ah, great catch, awesome!

PR #181, opened 2021-06-25
Fix typo

yangfl:
with the love from codespell


troglobit:
Awesome, thank you so much! :heart:

PR #199, opened 2021-08-29
Avoid using C++ key words

liuming50:
"protected" is a C++ key word, replace it with "is_protected" avoid
compiling failures with C++ compiler.


troglobit:
OK, I'll approve this.  But please be aware that Finit is a C project and only C compilers are officially supported.

PR #200, opened 2021-08-29
Dont hard-code nologin file

liuming50:
It's configurable in /etc/login.defs as NOLOGINS_FILE, avoid to hard
coding it here.


troglobit:
This was a bit confusing since you only moved the hard-coded string to another place, but I'm guessing you set `NOLOGIN` outside of the Finit build system?  A proper fix would be to use libpam to get the value of NOLOGINS_FILE, right?

I'd like to see the default use _PATH_NOLOGIN from /usr/include/paths.h.

liuming50:
> This was a bit confusing since you only moved the hard-coded string to another place, but I'm guessing you set `NOLOGIN` outside of the Finit build system? A proper fix would be to use libpam to get the value of NOLOGINS_FILE, right?
> 
> I'd like to see the default use _PATH_NOLOGIN from /usr/include/paths.h.

Hi,

Thanks for the review.

I agree with you so made this V2.

It does not refer to _PATH_NOLOGIN since I think it should not rely on any specific libc, and to support both PAM and non-PAM enabled systems, we now introduce one more config option. (BTW: there seems is no interface from PAM that we can query the nologin path, pam itself is using a default hard-coded nologin path which is /etc/nologin.)

troglobit:
Thanks for the update.  I'm not really happy about the inflation of configure options though, I spent the last release cycle cleaning up a lot of old such and I don't want to do that again.  A configure option should exist for *commonly* used options.  I have a hard time seeing `/etc/nologin` being such a common option.  It's better to keep such as possible to override as an external define, this is fine.

Also, I'm the maintainer, so let me repeat myself: "I'd like to see the default use _PATH_NOLOGIN from /usr/include/paths.h."

liuming50:
@troglobit thanks for the review, I have made a V3 based on your comment.

troglobit:
Thank you, much better! :-)

PR #201, opened 2021-08-29
Fix a memleak.

liuming50:


troglobit:
Ouch, nice catch, thanks!

PR #202, opened 2021-08-29
Introduce two more configure options

liuming50:
fastboot: once enabled, fsck would be skipped on filesystems listed in
          /etc/fstab.
fsckfix: once enabled, 'fsck -ys' would be run instead of 'fsck -a',
         and check the return value larger than 1 as a failure and goes
         to login console.

They are both disabled by default, no functional changes unless an end
user choose to enable them.

The idea derives from Sysvinit's "FASTBOOT" and "FSCKFIX".


troglobit:
Great idea!  If you feel up to it, this might be useful to be able to activate from the command line as well, from a boot loader.

Minor fixups only of coding style, added as follow-up commit.

troglobit:
Hand merged and pushed to master.

PR #203, opened 2021-08-30
Ensure all filesystems listed in /proc/mounts are unmounted

liuming50:
The iterator function getmntent() is not stable. It may skip
entries if the contents of the iterated file changes, which
is the case with /proc/mounts when filesystems are unmounted.
As a result, some filesystems were never unmounted.

To fix this, iteratation is now restarted after every
sucessful unmount.

### DIFF
--- a/src/mount.c
+++ b/src/mount.c
@@ -59,46 +59,54 @@ static int is_protected(char *dir)
 	return 0;
 }
 
-static struct mntent *iterator(char *fstab)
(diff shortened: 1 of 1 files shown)


troglobit:
Thanks, well spotted!

For future reference, I've added a follow-up commit that:
 1. Fixes the comment style; from C++ to C style
 2. Drops the extra `print()`, I'm assuming it was for debug

PR #204, opened 2021-09-21
Export libfinit library/headers

liuming50:
So finit could be linked by other programs, for instance, if they want
to implement their own specific plugins.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -14,6 +14,7 @@ AC_CONFIG_FILES([Makefile
 	man/Makefile
 	plugins/Makefile
 	src/Makefile
+	src/libfinit.pc
 	test/test.env
 	test/Makefile])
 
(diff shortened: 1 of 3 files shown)


liuming50:
Changes in patch set V2:

1 Export all header files from src

troglobit:
Hmm, I don't like this at all, sorry.  We've had this come up before and it's caused lots of headache whenever we change header files internally, with external plugins becoming incompatible in "interesting" ways. We still rely on it at $DAYJOB, so that's why there's still support for it, but it's something we're trying to phase out, since having 3rd party plugins in PID 1 is a huge problem (crashing PID 1 is no good) and potential security risk.

Exporting all headers like this would mean I have to take the full support if something breaks, which means I have to make sure to keep compatible structs/defines/functions/etc, and that's not something I'm overly keen on doing.  If we had full time staffing, or at least more active maintainers than just me, then I guess we could consider it.

Terribly sorry for the inconvenience, but I must decline this and close the PR.  I guess you'll have to maintain this as an out-of-tree patch on your side :-/

PR #207, opened 2021-09-21
service.c: set condition 'done' for oneshot tasks

liuming50:
We like to know when a oneshot task is done, so as to start the
services that depending on it, set a oneshot condition 'done' for
that case.

### DIFF
--- a/src/service.c
+++ b/src/service.c
@@ -1677,6 +1677,21 @@ static void svc_set_state(svc_t *svc, svc_state_t new)
 		service_timeout_cancel(svc);
 		service_timeout_after(svc, svc->killdelay, service_kill);
 	}
+
+	if (svc_is_runtask(svc)) {
+		char cond[MAX_COND_LEN];
+
+		snprintf(cond, sizeof(cond), "%s/%s/done", svc->type, svc->name);
+
+		/* create done condition when entering SVC_DONE_STATE. */
+		if (*state == SVC_DONE_STATE)
+			cond_set_oneshot(cond);
+
+		/* clear done condition when entering SVC_HALTED_STATE. */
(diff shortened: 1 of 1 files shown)


troglobit:
Interesting. It would be great if this also included an official way of accessing this information instead of poking around in the internals. Meaning, could you please add a way of inspecting this through `initctl status foo` as well?

Also, and possibly more importantly.  A one-shot task per spec. runs once per runlevel it's allowed to run in, so when changing runelevel the task is run once again.  Maybe this file should be removed for all tasks that are restarted?

liuming50:
> Interesting. It would be great if this also included an official way of accessing this information instead of poking around in the internals. Meaning, could you please add a way of inspecting this through `initctl status foo` as well?
> 
> Also, and possibly more importantly. A one-shot task per spec. runs once per runlevel it's allowed to run in, so when changing runelevel the task is run once again. Maybe this file should be removed for all tasks that are restarted?

Thanks for the review, will rework a V2.

troglobit:
Any progress on this?  I'm gearing up for release 4.2 very soon ... :)

troglobit:
Finit v4.2 was released during the weekend. Maybe we can get this into the next release cycle? :pray:

liuming50:
@troglobit sorry for the delayed feedback, we are currently being blocked by https://github.com/troglobit/finit/issues/216, so have not managed time work on this.

liuming50:
@troglobit 

Please kindly help me review this V2.

If I understood you correctly, when SM enters SVC_DONE_STATE, we set done condition, and when it enters SVC_HALTED_STATE, we clear the done condition. So I think the svc_set_state function is still the proper place to put the logic in.

troglobit:
Much better, thank you! :)

PR #208, opened 2021-09-21
service.c: add more parameters when starting services

liuming50:
It would be good if we can control the services more specifically with
the following parameters like:

restart:N   - how many times shall the service to be restarted on
              failures (<10), the default max is 10.
restart_tmo - the timeout of the restarting.
norestart   - dont restart on failures.
oncrash     - once all retrying also fail, how to deal with it,
              rebooting or ignoring the failures.

### DIFF
--- a/src/service.c
+++ b/src/service.c
@@ -26,6 +26,7 @@
 
 #include <ctype.h>		/* isblank() */
 #include <string.h>
+#include <sys/reboot.h>
 #include <sys/prctl.h>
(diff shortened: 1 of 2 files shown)


troglobit:
Awesome, we've been talking about adding this ourselves, very cool! <3

troglobit:
For future reference, it's always appreciated when contributors also add some basic documentation ;-)

PR #209, opened 2021-12-06
Fix filesystems not unmounted at shutdown

liuming50:
Instead of unconditionally waiting 2 seconds for processes to die,
check continuously for remaining processes, and break the loop when
none remain.

Also don't wait for the console -sh tty process.

Signed-off-by: Mathias Thore <mathias.thore@atlascopco.com>

### DIFF
--- a/src/helpers.c
+++ b/src/helpers.c
@@ -57,7 +57,7 @@ static pstyle_t progress_style = PROGRESS_DEFAULT;
 #define STATUS_CLASS {							\
 	CHOOSE(" OK ",  " OK ",  "\e[1;32m"),		/* Green  */	\
 	CHOOSE("FAIL",  "FAIL",  "\e[1;31m"),		/* Red    */	\
-	CHOOSE("WARN",  "WARN",  "\e[1;31m"),		/* Red    */	\
+	CHOOSE("WARN",  "WARN",  "\e[1;33m"),		/* Yellow */	\
 	CHOOSE(" \\/ ", " ⋯  ", "\e[1;33m"),		/* Yellow */	\
 }
 


troglobit:
In general I think this patch is a good idea.  I do however have several reservations; in reversed order:

 1. Why not wait for, and kill, the console -sh tty process? They definitely sit on a mounted filesystem (and may have several open files from others), that prevent us from unmounting cleanly
 2. Passing integers as `void *` in the callbacks don't work on all archs, pointers and integers are no longer guaranteed to be of the same length.  If you only ever want to pass an integer, use an  `int *`
 3. I'm really worried to see the change to `SCHED_RR` this late in the lifetime of Finit ... do we really need it?  Please motivate.
 4. Is the usleep() long enough for the kernel to schedule another task on an old Arm9 with no L2 cache?  We have several targets at $DAYJOB that use an old NPX i.MX27 that is soooo slow.

troglobit:
No reaction so far on my review for this?  I'm gearing up for release 4.2 very soon ... so it would be nice to know how to proceed with this one.

troglobit:
Finit v4.2 was released during the weekend. Maybe we can sort the review out for the next release cycle? :pray:

liuming50:
@troglobit sorry for the delayed feedback, we are currently being blocked by https://github.com/troglobit/finit/issues/216, so have not managed time work on this.

liuming50:
@troglobit Please kindly help me review the V2.

Please see my comments inline for your questions about V1:

> Why not wait for, and kill, the console -sh tty process? They definitely sit on a mounted filesystem (and may have several open files from others), that prevent us from unmounting cleanly

The original author Robert Andersson said he does not remember the exact reason, so I dropped it in V2, it might be some specific requirements from our system.

> Passing integers as void * in the callbacks don't work on all archs, pointers and integers are no longer guaranteed to be of the same length. If you only ever want to pass an integer, use an int *

Have changed as you suggested in V2.

> I'm really worried to see the change to SCHED_RR this late in the lifetime of Finit ... do we really need it? Please motivate.

Maybe I can explain this, this is basically for PREEMPT-RT kernel, finit during shutdown must run as a RT process to ensure it would not be scheduled out so as to satisfy threaded IRQs (which has RT priorities as 49) CPU requirement. This part of change should also work on non-preempt-rt kernel, so I kept it in V2, like to see your further comment.

> Is the usleep() long enough for the kernel to schedule another task on an old Arm9 with no L2 cache? We have several targets at $DAYJOB that use an old NPX i.MX27 that is soooo slow.

I am not 100% sure about the answer, but we do deploy finit to NXP imx27 boards (maybe thousands of them), I have not seen errors about this change.

PR #219, opened 2022-02-06
Fix filesystems not unmounted at shutdown

liuming50:
Instead of unconditionally waiting 2 seconds for processes to die,
check continuously for remaining processes, and break the loop when
none remain.

Turn PID 1 to a RT process with highest priority 99 during shutdown,
this ensures it would not be preempted by other RT processes.

Signed-off-by: Mathias Thore <mathias.thore@atlascopco.com>

### DIFF
--- a/src/sig.c
+++ b/src/sig.c
@@ -71,6 +71,7 @@
 
 #include <dirent.h>
 #include <string.h>		/* strerror() */
+#include <sched.h>
 #include <sys/reboot.h>
 #include <sys/wait.h>
 #ifdef _LIBITE_LITE
@@ -172,7 +173,7 @@ void unmount_regular(void);
  *
  * https://www.freedesktop.org/wiki/Software/systemd/RootStorageDaemons/
  */
-void do_kill(int signo)
+void do_iterate_proc(int (*callback)(int, int *), int *context)
 {
 	DIR *dirp;
 
@@ -191,6 +192,8 @@ void do_kill(int signo)
 			pid = atoi(d->d_name);
 			if (!pid)
 				continue;
+			if (pid == 1)
+				continue;
 
 			snprintf(file, sizeof(file), "/proc/%s/cmdline", d->d_name);
 			fp = fopen(file, "r");
@@ -203,7 +206,10 @@ void do_kill(int signo)
 				else if (file[0] == '@')
 					_d("Skipping %s ...", &file[1]);
 				else
-					kill(pid, signo);
+					if (callback(pid, context)) {
+						print(0, "PID %d is still alive (%s)", pid, file);
+						break;
+					}
 			}
 			fclose(fp);
 		}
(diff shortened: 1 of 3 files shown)


liuming50:
Hi, @troglobit 

I mistakenly closed https://github.com/troglobit/finit/pull/209, I created this new MR, please help me review it.

Paste the answers to your question here:

> Why not wait for, and kill, the console -sh tty process? They definitely sit on a mounted filesystem (and may have several open files from others), that prevent us from unmounting cleanly

The original author Robert Andersson said he does not remember the exact reason, so I dropped it in V2, it might be some specific requirements from our system.

> Passing integers as void * in the callbacks don't work on all archs, pointers and integers are no longer guaranteed to be of the same length. If you only ever want to pass an integer, use an int *

Have changed as you suggested in this MR.

> I'm really worried to see the change to SCHED_RR this late in the lifetime of Finit ... do we really need it? Please motivate.

Maybe I can explain this, this is basically for PREEMPT-RT kernel, finit during shutdown must run as a RT process to ensure it would not be scheduled out, so it must run at the highest priority 99. This part of change should also work on non-preempt-rt kernel.

> Is the usleep() long enough for the kernel to schedule another task on an old Arm9 with no L2 cache? We have several targets at $DAYJOB that use an old NPX i.MX27 that is soooo slow.

I am not 100% sure about the answer, but we do deploy finit to NXP imx27 boards (maybe thousands of them), I have not seen error report about this change, so 2500 USECS should be long enough I guess.

src/util.c
```diff
@@ -297,6 +297,19 @@ void do_sleep(unsigned int sec)
...
+void do_usleep(unsigned int usec)
+{
+	struct timespec deadline;
+	clock_gettime(CLOCK_MONOTONIC, &deadline);
+
+	deadline.tv_nsec += usec * 1000;
+	deadline.tv_sec += deadline.tv_nsec / 1000000000;
+	deadline.tv_nsec = deadline.tv_nsec % 1000000000;
+
+	while (clock_nanosleep(CLOCK_MONOTONIC, TIMER_ABSTIME, &deadline, NULL) < 0 && errno == EINTR)
```
troglobit, line 309: The clock_nanosleep() API seems to return positive error codes, never negative ones.  Yes, I was also surprised.

src/util.c
```diff
@@ -297,6 +297,19 @@ void do_sleep(unsigned int sec)
 		;
 }
 
+void do_usleep(unsigned int usec)
+{
+	struct timespec deadline;
+	clock_gettime(CLOCK_MONOTONIC, &deadline);
```
troglobit, line 303: Nitpick, should be an empty line after declarations.

src/sig.c
```diff
@@ -203,16 +206,60 @@ void do_kill(int signo)
...
+	}
+	while (has_proc && iterations > 0);
+
+	return has_proc;
+}
+
 void do_shutdown(shutop_t op)
 {
+	struct sched_param sched_param;
+	sched_param.sched_priority = 99;
```
troglobit, line 258: Nitpick, empty line after declarations, or `sched_param = { .sched_priority = 99 };`

src/sig.c
```diff
@@ -203,16 +206,60 @@ void do_kill(int signo)
...
+
+	/* PID 1 run as RR process with highest priority */
+	sched_setscheduler(1, SCHED_RR, &sched_param);
```
troglobit, line 261: The comment is useless as-is.  It states in text what the next line does ... which can be read from the code itself. Instead, use comments to explain *why* a code path is needed. Here's a good place to add the reasoning from the discussion.  E.g., *"On a PREEMPT-RT system, Finit must run as the highest prioritized RT process to ensure it completes the shutdown sequence."*

troglobit (changes requested):
OK, I think I understand a bit better now why you need this change.  I'm still not overly happy with the change, but I cannot see a better way to solve this at the moment.  

Please address my review comments, in particular the `clock_nanosleep()` issue, then I'll merge it.  If you're quick it'll make it into the next release :)

src/util.c
liuming50: Arha, did not notice that, nice catch, will fix it in V2.

src/sig.c
liuming50: Will do that!

src/sig.c
liuming50: @troglobit Good point, will do that.

src/util.c
liuming50: Will change that in V2!

liuming50:
@troglobit Please kindly help me review the V2.

Have fixed all issues as you suggested in V1.

troglobit (approved):
Looks great, just as we discussed! :)

PR #220, opened 2022-02-07
service.c: fix a regression

liuming50:
A regression was introduced by commit f0f358a1:
[ service.c: set/clear condition 'done' for run tasks ]

svc->type is a integer but mistakenly being used as string, which will
cause crash.

### DIFF
--- a/src/service.c
+++ b/src/service.c
@@ -1709,7 +1709,7 @@ static void svc_set_state(svc_t *svc, svc_state_t new)
 	if (svc_is_runtask(svc)) {
 		char cond[MAX_COND_LEN];
 
-		snprintf(cond, sizeof(cond), "%s/%s/done", svc->type, svc->name);
+		snprintf(cond, sizeof(cond), "%s/%s/done", svc_typestr(svc), svc->name);
 
 		/* create done condition when entering SVC_DONE_STATE. */
 		if (*state == SVC_DONE_STATE)
--- a/src/initctl.c
+++ b/src/initctl.c
@@ -292,7 +292,7 @@ static int dump_one_cond(const char *fpath, const struct stat *sb, int tflag, st
 	if (strncmp("pid/", cond, 4) == 0) {
 		svc_t *svc;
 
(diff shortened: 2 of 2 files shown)


src/service.c
```diff
@@ -1709,7 +1709,20 @@ static void svc_set_state(svc_t *svc, svc_state_t new)
...
+		switch (svc->type) {
+		case SVC_TYPE_TASK:
+			snprintf(cond, sizeof(cond), "task/%s/done", svc->name);
+			break;
+		case SVC_TYPE_RUN:
+			snprintf(cond, sizeof(cond), "run/%s/done", svc->name);
+			break;
+		case SVC_TYPE_SYSV:
+			snprintf(cond, sizeof(cond), "sysv/%s/done", svc->name);
+			break;
```
troglobit, line 1721: Good catch, I didn't spot this in my review, sorry about that!

In general I try to avoid spreading out hard coded constants.  Could you please change the above to:
```C
  case SVC_TYPE_TASK:
  case SVC_TYPE_RUN:
  case SVC_TYPE_SYSV:
     snprintf(cond, sizeof(cond), "%s/%s/done", svc_typestr(svc), svc->name);
     break;
```

troglobit (changes requested):

src/service.c
liuming50: @troglobit 

Arha, good suggestion, then maybe the switch/case could be dropped as well, just use "svc_typestr(svc)", will send a V2.

liuming50:
@troglobit Changes in V2:

1 Adjusted as you suggested in V1.

2 Fix another inconsistent coding style.

liuming50:
@troglobit Verified the crash fixed on a NXP imx27 and imx51 board.

troglobit:
Thanks, looks good to me! :)

src/service.c
troglobit: Well maybe, it served as a neat filter -- only task/run/sysv can create the done condition.  But either way is fine with me :)

Issue #225, opened 2022-02-11
initctl signal <name> <signal> for send process signals


hongkongkiwi:
There are times where I need to send a custom signal directly to a running process.

For example, udhcpc can receive a custom user signal to release or renew an ip if running as a daemon.

It would be great if there was a command in initctl to do this. I am thinking something like:
`initctl signal <name> <signal>`

For example
`initctl signal udhcpc SIGUSR1`

I do have a workaround but it is not very elegant:
`initctl -b -p status udhcpc| grep '\s*PID\s:' | sed 's|.*: ||' | xargs kill -s SIGUSR1`

troglobit:
Neat idea, I'm using something similar myself right now! :)

jorgensigvardsson:
So, basically, this issue is about locating the correct PID for the currently executing service, and send the requested signal to it.

Shouldn't this be restricted in some way? Normally you can't do this to any other process than your own:

> The user running initctl may not be root so we need the signaling to be done from inside Finit

Do you mean that this function is free for all? No group membership required, or finit config specifying which users (or groups) that may do this? Globally or per service (or both)?

troglobit:
Just to clarify, it's to send a custom signal to a run/task/service *managed* by Finit.  Not unlike the start/stop/restart commands. Maybe that info got lost.  (For other processes there's `kill(1)`.)

Yes, it's currently restricted by 1) the execute permission flags on `initctl`, 2) the socket file in `/var/run`.  By default, Finit installs with root permissions to use any of the initctl commands -- however, a sysadmin or a Linux distribution may choose to have initctl suid root, or change the group permission flags.  E.g., all members of the `wheel` group have permission to manage the system with initctl.  Meaning, they are not root and cannot use `kill(1)` to send signals, but they can use `initctl signal`.  No other mechanisms to restrict access exist at present, and unlike systemd, Finit does not dictate any policy to distributions.

Sorry for the confusion my comment may have caused.  I sometimes forget to mention all the specific use-cases, and the details surrounding them, that exist.

jorgensigvardsson:
Ok, `initctl` extended: https://github.com/jorgensigvardsson/finit/commit/bc1bf579f789157c6649999d2a23aeb9bb7d9485

I have extended the command parsing with a new callback `cb_multiarg(int argc, char **argv)`, because I needed two args for the same function. Could not use the same multi level command hack as `cond` uses.

[...]

PR #229, opened 2022-02-13
initctl signal

jorgensigvardsson:
Implements the feature requested in #225.

`initctl` has been changed in the following way:
- it has been augmented with a new command: `signal`. The command requires two arguments: the service to signal, and the signal itself.
- the signal can be a complete signal name such as `SIGHUP`, or a short name such as `HUP`, or the signal number such as `1`
- man page for `initctl` has been updated accordingly

Please note that internal APIs have been changed:
- `initctl` has been modified such that it can implement multi argument commands. Please see `cb_multiarg` of `struct cmd` in `initctl.c`. Changes are backwards compatible with existing code (old command handlers have not been changed).
- `svc_parse_jobstr` from `svc.{h,c}` has been extended so that it can pass on user data to any of its callbacks. The reasoning is that I wanted to reuse a lot of existing code that did most of what's needed. The needed functionality was to pass the actual signal number to the callback: this is done using the `user_data` parameter.


troglobit:
Build and test pass! \o/

I'll have to postpone the review for tomorrow, it got a bit too late tonight, sorry.  From a *very* quick look it seems most of my feedback will be around grammar, commit message style, and such trivial matters -- which I know will be important for you to learn going forward in other upstream projects as well.

jorgensigvardsson:
Sorry about the commit spam to this PR. I have just learned that one should definitely `make distclean` et al, and build everything from scratch. Did some changes to test scripts, ran `make check` - everything checked out. Committed and pushed, and still it failed in the github workflows. Only when I cleaned and built everything from scratch did I notice the exact same problems.

troglobit:
@jorgensigvardsson there's a actually `make distcheck` build target for that, very handy and what I call in my `release` target :)

troglobit:
Sorry for checking back so late on this, but as you know, it was crazy buys at $DAYJOB this past week!  :disappointed: 

The PR looks great, only really had one or two minor nitpicks, but I'll fix up manually and merge anyway. If you're curious, check the follow-up commits later today.

The stuff that sticks out, however, are the commit messages.  I'll just mention a thing or two about it here, for future reference:

   * When fixing a bug/feature in the issue tracker, say `"Fix #225: initctl support for sending signals to services"`.  So when the maintainer merges to master, GitHub will automatically close issue `#225`. I'll add it now to the merge commit instead. The PR text itself, which is similar to a mailing list cover letter will be lost and only stored with GitHub, so some text should be in the commit that ties it all together.  I.e., `"This patch implements the `initctl signal` command using the recently added API for signaling ... "`
   * Always use present tense, i.e., `"Add implementation for ..."`

PR #232, opened 2022-02-26
service.c: create conditions according to runtask exit status

liuming50:
In commit f0f358a13:
[ service.c: set/clear condition 'done' for run tasks ]

a runtask done condition would be set/cleared when entering DONE/HALTED
states, but it did not cover all the user cases, for instance, sometimes
an end user may want to know if a runtask has finished sucessfully or
to decide what to do on its failures.

So we now change the conditions to: tsktype/tskname/success and
tsktype/tskname/failure.

And this change not only applies to run/task types, but also applies to
sysv type, in case it fails, a sysv/tskname/failure condition would be
set.

### DIFF
--- a/src/service.c
+++ b/src/service.c
@@ -628,10 +628,13 @@ static int service_start(svc_t *svc)
 	switch (svc->type) {
 	case SVC_TYPE_RUN:
 		svc->status = complete(svc->cmd, pid);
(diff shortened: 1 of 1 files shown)


troglobit:
Nicely done.  No comments from me on this one :-)

(Sorry for taking so long to respond! In the middle of changing jobs)

PR #233, opened 2022-03-03
initctl.c: fix a return value of run tasks

liuming50:
When a run task is started with svc->started = 1, it should be
considered started successfully so it should return 0.

### DIFF
--- a/src/initctl.c
+++ b/src/initctl.c
@@ -791,8 +791,15 @@ static int show_status(char *arg)
 		if (!svc)
 			return 255;
 
-		if (quiet)
-			return svc->state != SVC_RUNNING_STATE;
+		if (quiet) {
+			if (svc_is_runtask(svc))
+				if (svc->started)
+					return 0;
+				else
+					return 1;
+			else
+				return svc->state != SVC_RUNNING_STATE;
+		}
 
 		pidfn = svc->pidfile;
 		if (pidfn[0] == '!')
--- a/src/svc.h
+++ b/src/svc.h
@@ -332,7 +332,10 @@ static inline char *svc_status(svc_t *svc)
 		return "unknown";
 
 	case SVC_DONE_STATE:
-		return "done";
+		if (svc->started)
+			return "done";
+		else
+			return "failed";
 
 	case SVC_STOPPING_STATE:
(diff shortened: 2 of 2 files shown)


src/svc.h
```diff
@@ -332,7 +332,10 @@ static inline char *svc_status(svc_t *svc)
 		return "unknown";
 
 	case SVC_DONE_STATE:
-		return "done";
+		if (svc->started)
+			return "succeeded";
+		else
+			return "failed";
```
troglobit, line 338: I was tempted to merge this as-is, but I'm afraid I cannot approve this particular part, because there are tools out there (e.g. at $DAYJOB) that parse this output, expecting the string "done", and would break with this change.

If you change back to "done", keeping "failed", and restore the formatting above, then I can merge it.

src/service.c
```diff
@@ -1749,7 +1749,7 @@ int service_step(svc_t *svc)
 	old_state = svc->state;
 	enabled = svc_enabled(svc);
 
-	_d("%20s(%4d): %8s %3sabled/%-7s cond:%-4s", svc->cmd, svc->pid,
+	_d("%20s(%4d): %10s %3sabled/%-7s cond:%-4s", svc->cmd, svc->pid,
```
troglobit, line 1752: (this)

src/initctl.c
```diff
@@ -738,7 +738,7 @@ static char *status(svc_t *svc, int full)
 		color = NULL;
 
 	if (!full)
-		snprintf(buf, sizeof(buf), "%-8.8s", s);
+		snprintf(buf, sizeof(buf), "%-10.10s", s);
```
troglobit, line 741: (this)

troglobit (changes requested):

src/svc.h
liuming50: @troglobit sure, I can do that.

liuming50:
@troglobit Changes in patch set V2:

1 Changes combined to one commit.
2 Change the status string to 'done' as you suggested.

troglobit:
Awesome, thank you! <3

troglobit (approved):

PR #293, opened 2022-06-23
Support for coldplugging more devices

wkz:
As it turns out, not all buses in Linux will add `modalias` attributes to their devices in sysfs. One notable exception are MDIO buses. The plugin's scan routine would thus not pick them up.

Fortunately, the information is always available in the device's `uevent` attribute.

Therefore, default to looking for modaliases in the device's `uevent` attribute instead, falling back to `modalias` only if `uevent` is not availble (this should never happen).


troglobit:
Awesome work on this, thank you! :heart:

PR #337, opened 2023-01-06
Fix AUTO_RELOAD #define typo

JackNewman12:
It looks like the makefiles are creating a `AUTO_RELOAD` define, but the code is looking for `ENABLE_AUTO_RELOAD`. Making this change fixes the auto-reload feature for me. 

I also noticed that `conf.c do_change()` seems to ignore deleted files. 
Commenting out this checks causes a deleted or moved .conf to correctly remove a service. Not sure why this check is here so I'm afraid I might be missing some edge-cases.

```diff
 static int do_change(char *dir, char *name, uint32_t mask) 
 { 
 	char fn[strlen(dir) + strlen(name) + 2]; 
 	struct conf_change *node; 
  
 	paste(fn, sizeof(fn), dir, name); 
 	dbg("path: %s mask: %08x", fn, mask); 
  
 	node = conf_find(fn); 
- 	if (mask & (IN_DELETE | IN_MOVED_FROM)) { 
- 		drop_change(node); 
- 		return 0; 
- 	}
	if (node) {
		dbg("Event already registered for %s ...", name);
		return 0;
	}
```

### DIFF
--- a/src/cgroup.c
+++ b/src/cgroup.c
@@ -287,7 +287,7 @@ static void cgroup_events_cb(uev_t *w, void *arg, int events)
 		cgroup_handle_event(iwp->path, ev->mask);
 	}
 
-#ifdef ENABLE_AUTO_RELOAD
+#ifdef AUTO_RELOAD
 	if (conf_any_change())
 		service_reload_dynamic();
 #endif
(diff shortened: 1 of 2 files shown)


troglobit:
Ah thanks, a too hasty refactor by me.  Great fix!

About the second part, looks weird and doesn't have a comment to *why* so I'm inclined to drop it.  But I'll get to you on that, creating a separate issue for it.

PR #373, opened 2023-10-05
Set hostname as soon as possible

liuming50:
We find it's useful to get know hostname before plugins get initialized, since some plugins might want to do some machine specific settings.

This change mainly aims to satisfy that requirement, to set hostname as soon as possible in main function.

### DIFF
--- a/src/finit.c
+++ b/src/finit.c
@@ -648,6 +648,11 @@ int main(int argc, char *argv[])
 		rescue = sulogin(0);
 #endif
 
+	/* Set hostname as soon as possible, so it could be accessed
+	 * in plugins.
+	 */
+	set_hostname(&hostname);
+
 	/*
 	 * Load plugins early, the first hook is in banner(), so we
 	 * need plugins loaded before calling it.


troglobit:
With the current state of Finit, this change mean we'd set hostname twice during bootstrap, with potentially different names since at the point of your change the file `/etc/finit.conf` has not yet been read.

liuming50:
> With the current state of Finit, this change mean we'd set hostname twice during bootstrap, with potentially different names since at the point of your change the file `/etc/finit.conf` has not yet been read.

Yes, I am also aware of that, it will invalidate the DEFHOST or the setting from /etc/finit.conf, though it would not harm since at least the hostname would be set to /etc/hostname any way if it exists.

Do you have any ideas about how to support this kind of requirement? Or do you think it's a valid requirement to allow hostname to be accessed during plugins initializing?

troglobit:
Well, my gut reaction is to not accept this PR since I think we should not change hostname until we've had the chance to evaluate the configuration.

For one of my own use-cases our `/etc` is read-only in the boot image.  We then apply an OverlayFS (tmpfs) so we can write/change files in `/etc`.  As the last stage of bootstrap we generate `/etc/hostname` from a central file (startup-config -> NETCONF (sysrepo) => `/etc/hostname`).  So for us this PR would mean potentially three changes at boot.

Also, I do not see why a plugin would want to (or need to) know the (potential) hostname this early.  Why do you have this requirement, does your plugins extract a device role or something from the hostname, e.g. *"door-controller1"* or similar?  Can't they read out that information from `/etc/hostname` themselves in that case?

Maybe unrelated, but worth mentioning since I've just made the change right before this last release candidate for v4.5.  The native dbus plugin registers the dbus service with Finit early.  This was a problem in one of our setups so I've added a "delayed" registration where plugins can ask Finit to generate a `foo.conf` that it loads after all critical services have been loaded.  So dbus.so now creates `/run/finit/system/dbus.conf`, and if that service needed hostname it would be correctly set at that point in the execution.  For more info, in case you're interested, see 5d703fd.

Issue #374, opened 2023-10-15
Automount of /run and tmpfiles.d creating /run/utmp lacking documentation


az143:
situation: glibc 2.31, mdev instead of udev, no hook scripts or other helping hands that do anything with /run, 
no /run in /etc/fstab.

in that case finit mounts /run as tmpfs, then continues to  utmp_set_boot() and set_utent() - but nothing creates a missing utmp file, so setutent() fails so no utmp file, no boot entry therein  and thus tools like runlevel & co fail to determine the runlevel.
for comparison: sysvinit creates that file by script, and the documentation for setutent() doesn't say anything about utmp file 
autocreation, so i think finit is at fault here and should create it. 

two-line fix to follow via PR in a few minutes.

troglobit:
This should be handled already by the new tmpfiles.d support[^1] in the bootmisc plugin, which runs after all filesystems are mounted, including any missing `/run` set up by finit with a tmpfs.

Since the canonical path is `/var/run/utmp`, I have to ask, do you have the compat symlink `/var/run -> /run` on your system?

[^1]: Static files, directories, etc. needed by the system are being migrated to using the tmpfiles.d support instead of hard-coding mkdir/touch.

az143:
this is debian on an embedded system, so, yes: /var/run points to /run. support for tmpfiles.d doesn't help as this does not  have anything systemd-ish on it.  personally i'd have expected finit to not require systemd components for normal operation.

troglobit:
Finit has its own bare bones tmpfiles.d.

az143:
i must have screwed up my installation... which i seem to have; the tmpfiles.d files are not where they should be. 
from reading the docs, i didn't realise that the bootmisc plugin implements that aspect of finit. ok to leave this open until i submit a PR for expanding that bit of the documentation a little?

troglobit:
> i must have screwed up my installation... which i seem to have; the tmpfiles.d files are not where they should be. 
> 
> from reading the docs, i didn't realise that the bootmisc plugin implements that aspect of finit. ok to leave this open until i submit a PR for expanding that bit of the documentation a little?

The docs could use a real overhaul/split, I've planned this in for 5.0. In the meantime anyone is welcome to improve on the sad state things are. Thanks

troglobit:
Updated `doc/bootstrap.md` and `doc/plugins.md` slightly for this release.

PR #375, opened 2023-10-15
fix missing creation of utmp file

az143:
issue #374


troglobit:
Closing, verified working on upcoming v4.5 release with auto-mount of `/run` and tmpfiles.d in  `bootmisc.so` creating an empty `/var/run/utmp`.

Issue #377, opened 2023-10-23
service environment variables are not expanded when sourced


az143:
when setting up an `/etc/default/somesvc`, it is currently not possible to reference  variables defined earlier, because 
source_env() only strips undesirable bits before calling setenv().

for example, this construction doesn't work:
```
RUNDIR=/var/run/somesvc
DAEMON_ARGS=--workdir $RUNDIR --other-args...
```
i think that running wordexp() before the setenv() calls would remove this limitation; if that's not true, 
then the limitation should be mentioned in doc/config.md and man/finit.conf.5.

az143:
see pr #379

troglobit:
A `wordexp()` of the value (in key=value) is probably the way to go here.  It fits with the principle of least surprise, does not leak to other processes, and would also keep the documentation to a minimum.

PR #379, opened 2023-10-24
documentation fix for issue #377

az143:
...if the wordexp() approach is not taken, then the docs should spell out that service env variable references aren't expanded.


troglobit:
I prefer the `wordexp()` suggestion, as per my comment in https://github.com/troglobit/finit/issues/377#issuecomment-1777551886, closing.

Issue #380, opened 2023-10-24
Does finit start rc.local and gettys too early?


az143:
finit's start sequence does not quite conform to what the startup.md and other docs indicate:
rc.local is run as the second thing in runlevel 2 (just after getty setup), not as the last thing
as it should be. finit version is 4.5-rc5.

here is an excerpt of the finit's syslog entries:
```
Oct 23 13:45:40 testunit finit[1]: Starting coldplug[171]
Oct 23 13:45:41 testunit finit[1]: Starting bbb-hw[233]
...some more configured S tasks/services
Oct 23 13:45:44 testunit finit[1]: Debian GNU/Linux 10 (buster), entering runlevel 2
Oct 23 13:45:44 testunit finit[1]: Starting tty:S0[414]
Oct 23 13:45:48 testunit finit[1]: Starting redis-server[686]
...some more runlevel 2 services
```
for debugging purposes, my finit has been changed to log all progress (not just runlevel S;
i commented enable_progress(0) in src/sm.c around line 218), and to not
show a 'press enter to active this console' text
and here is the corresponding boot console output:
```
● ● ●  Debian GNU/Linux 10 (buster) ════════════
═══════════════════════
[ OK ] Checking filesystem /dev/mmcblk1p1
...
[ OK ] Remounting / as read-write
[ OK ] Mounting filesystems from /etc/fstab
[ OK ] Seeding random number generator
...
[ OK ] Starting mdev daemon
[ OK ] mdev cold plug run
[ OK ] BBB Hardware Init
...
[ OK ] Bringing up network interfaces ...
[ OK ] Getty on /dev/ttyS0
[ OK ] Calling /etc/rc.local in the background
[ OK ] Starting Redis Server
```

this clearly shows that the getty and rc.local are started before
any of the runlevel 2 services, not afterwards (as bootstrap.md cand the comments in sm.c laim, and as i'd have expected
coming from sysvinit where rc.local has a ```# Required-Start:    $all``` to make it come last).

the finit.d service definitions on that unit are very plain and boring;
the relevant excerpt:

```
tty [12345] /dev/ttyS0 noclear 115200 vt220
service [S12345] pid:/run/mdev.pid mdev -df -- mdev daemon
run [S] <service/mdev/ready> /sbin/coldplug -- mdev cold plug run
run [S] <service/mdev/ready,run/coldplug/success> /etc/init.d/bbb-hw -- BBB Hardware Init
[...]

PR #381, opened 2023-10-24
wishlist: option for progress output on all runlevels

az143:
i found debugging of finit's behaviour at and after the transition to runlevel 2 a bit troublesome, because all progress output stops when finit leaves runlevel S.

this PR adds the configure option ```--enable-allprogress```, which when enabled makes finit print progress outputs to the console on all runlevels instead of just S.

if that change is deemed too ugly/intrusive, then i'd suggest reviewing and fixing up the ```SM_BOOTSTRAP_WAIT_STATE``` logic in sm_step(), because that currently kills progress outputs before any level 2 services are started and that behaviour doesn't correspond to what man/finit.8 says about progress output, which is
```
When all services and run/tasks have been started, the console progress
is disabled and all configured getty services are started.
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -63,6 +63,10 @@ AC_ARG_ENABLE(fastboot,
         AS_HELP_STRING([--enable-fastboot], [Skip fsck check on filesystems listed in /etc/fstab]),,[
 	enable_fastboot=no])
 
+AC_ARG_ENABLE(allprogress,
+        AS_HELP_STRING([--enable-allprogress], [Progress output on all runlevels, not just S]),,[
+	enable_allprogress=no])
+
 AC_ARG_ENABLE(fsckfix,
         AS_HELP_STRING([--enable-fsckfix], [Run fsck fix mode (options: -yf) on filesystems listed in /etc/fstab]),,[
 	enable_fsckfix=no])
@@ -179,6 +183,10 @@ AS_IF([test "x$enable_kernel_cmdline" = "xyes"], [
 
 AS_IF([test "x$enable_fastboot" = "xyes"], [
 	AC_DEFINE(FAST_BOOT, 1, [Skip fsck check on filesystems listed in /etc/fstab])])
+
+AS_IF([test "x$enable_allprogress" = "xyes"], [
+	AC_DEFINE(ALL_PROGRESS, 1, [Progress output on all runlevels, not just S])])
+
 
 AS_IF([test "x$enable_fsckfix" = "xyes"], [
 	AC_DEFINE(FSCK_FIX, 1, [Run fsck fix mode (options: -yf) on filesystems listed in /etc/fstab])])
--- a/src/sm.c
+++ b/src/sm.c
@@ -218,7 +218,7 @@ restart:
 		plugin_run_hooks(HOOK_SYSTEM_UP);
 		service_step_all(SVC_TYPE_ANY);
 
-#ifndef ALLPROGRESS
+#ifndef ALL_PROGRESS
 		/* Disable progress output at normal runtime */
 		enable_progress(0);
 #endif
--- a/src/sm.c
+++ b/src/sm.c
@@ -218,8 +218,10 @@ restart:
 		plugin_run_hooks(HOOK_SYSTEM_UP);
 		service_step_all(SVC_TYPE_ANY);
 
+#ifndef ALLPROGRESS
 		/* Disable progress output at normal runtime */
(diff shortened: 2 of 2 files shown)


troglobit:
Many optimizations of the bootstrap process have been made in the 4.x release series and the documentation is unfortunately not up to date.  In this case it's wrong, misleading, or simply lagging behind. Progress is disabled "when all services and run/tasks _in runlevel S_ have been started".  Please don't expect the documentation to be a blueprint.

I'm going to close this since it's not something I want to maintain in this project.

As a final note, and review comment, in this project we appreciate squashing and force-pushing fixes like your last one; "fix stupid ..."

az143:
On Tue, 24 Oct 2023 08:59:45 -0700, Joachim Wiberg writes:
>In this case it's wrong, misleading, or simply lagging behind.
>Progress is disabled "when all services and run/tasks _in runlevel S_
>have been started".

if that's the way you want it to behave, then i think it needs to
be noted somewhere: sudden silence violates the principle of least
surprise quite a bit. (sysvinit doesn't stop progress outputs, dunno about
systemd or others.)

besides, the behaviour you describe above doesn't quite line up
with the docs _and_ the code:  bootstrap.md says step 26 hook_system_up,
and sm.c around line 220 calls enable_progress(0) after
plugin_run_hooks(HOOK_SYSTEM_UP). this implies to me that all of that
happens substantially later than step 23, switch to configured runlevel.

(i'm still a bit suspicious of a sequencing gotcha somewhere
in this region of post-S/pre-N handling, what with #380 which sortakinda
overlaps this.)

>Please don't expect the documentation to be a blueprint.

i'm not, but sm.c's ordering should be.

>I'm going to close this since it's not something I want to maintain in
>this project.

that's fine, i've got no problem with that decision. (that the
rejected PR exist and the closed problem report is findable by others
is good enough for me.)

>As a final note, and review comment, in this project we appreciate squashing an
>d force-pushing fixes like your last one; "fix stupid ..."

yes, sorry - i was a bit rushed and screwed that one up a little.

regards
az


-- 
Alexander Zangerl + GPG Key 2FCCF66BB963BD5F + https://snafu.priv.at/
grep..grep..grep.. (frog with UNIX stuck in it's throat)

troglobit:
> Alexander Zangerl writes:
> > On Tue, 24 Oct 2023 08:59:45 -0700, Joachim Wiberg writes:
> > [docs] In this case it's wrong, misleading, or simply lagging behind. Progress is disabled "when all services and run/tasks _in runlevel S_ have been started".
> if that's the way you want it to behave, then i think it needs to be noted somewhere:
> sudden silence violates the principle of least surprise quite a bit. (sysvinit doesn't stop
> progress outputs, dunno about systemd or others.)

Again, this is *not* SysV init, nor does it try to be a replacement for SysV init.  

Why do you re-use my own phrase against me, do you somehow think that will win you points?  Let me tell you what it does; it makes me feel tired and weary about this whole conversation I've spent far too many of my spare time hours on -- to help you.  Instead you press on like you are in your right to demand something?

> besides, the behaviour you describe above doesn't quite line up with the docs _and_ the code: bootstrap.md says step 26 hook_system_up, and sm.c around line 220 calls enable_progress(0) after plugin_run_hooks(HOOK_SYSTEM_UP).  this implies to me that all of that happens substantially later than step 23, switch to configured runlevel. (i'm still a bit suspicious of a sequencing gotcha somewhere in this region of post-S/pre-N handling, what with #380 which sortakinda overlaps this.)

This is getting out of hand, what if I remove all documentation, would that be better?!  I think I'll start by removing `doc/bootstrap.md`, it only exists because a customer wanted to know the bootstrap in v3.x ... it's out of date and I don't want to maintain that file anyway, and you keeping citing it back to me as some sort of disapproval of design choices and changes in _my own_ project is "sortakinda" wearing me down, not giving me any positive feedback *at all* ... If you have other expectations, the LICENSE file allows you to fork the software and do something better yourself.

> > Please don't expect the documentation to be a blueprint.
> i'm not, but sm.c's ordering should be.

The ordering in `src/sm.c` is basically "don't launch rc.local (in the background) until we know all run/task/service in runlevel S have completed".  If I change that to find a few 100 msec, then I've changed the blueprint ... what's your point?

> > I'm going to close this since it's not something I want to maintain in this project.
> that's fine, i've got no problem with that decision. (that the rejected PR exist and the closed problem report is findable by others is good enough for me.)

The arguing seems to indicate you do have a problem with my decision though.

PR #388, opened 2023-12-16
Fix log child exiting with debug disabled

soggyfries:
When debug is disabled an empty string `""` is sent as an arg to the logger process specified with `log`, both `logit` and `logger` interpret this as a log message and exit after emitting a single message containing the following parameters.

### DIFF
--- a/src/service.c
+++ b/src/service.c
@@ -330,14 +330,15 @@ static int lredirect(svc_t *svc)
 
 				snprintf(rot, sizeof(rot), "%d:%d", logfile_size_max, logfile_count_max);
 				snprintf(pid, sizeof(pid), "%d", svc_pid);
-				execlp("logger", "logger", "-f", svc->log.file, "-b", "-t", tag, "-p", prio, debug ? "-s" : "", "-I", pid, "-r", rot, NULL);
+				execlp("logger", "logger", "-f", svc->log.file, "-b", "-t", tag, "-p", prio, "-I", pid, "-r", rot, debug ? "-s" : NULL, NULL);
 			} else {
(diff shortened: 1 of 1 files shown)


src/service.c
```diff
@@ -330,14 +330,21 @@ static int lredirect(svc_t *svc)
 
 				snprintf(rot, sizeof(rot), "%d:%d", logfile_size_max, logfile_count_max);
 				snprintf(pid, sizeof(pid), "%d", svc_pid);
-				execlp("logger", "logger", "-f", svc->log.file, "-b", "-t", tag, "-p", prio, debug ? "-s" : "", "-I", pid, "-r", rot, NULL);
+				if (debug)
+					execlp("logger", "logger", "-f", svc->log.file, "-b", "-t", tag, "-p", prio, "-s", "-I", pid, "-r", rot, NULL);
+				else
+					execlp("logger", "logger", "-f", svc->log.file, "-b", "-t", tag, "-p", prio, "-I", pid, "-r", rot, NULL);
```
troglobit, line 336: Great find!

However, isn't a better solution to just move the optional `-s` to the end?  E.g.,

```C
execlp("logger", "logger", "-f", svc->log.file, "-b", "-t", tag, "-p", prio, "-I", pid, "-r", rot, debug ? "-s" : NULL, NULL);
```

Please change to that for all these I'll be more than happy to merge the PR.

troglobit (changes requested):

src/service.c
soggyfries: Didn't even think of that, much simpler. Changed.

troglobit (approved):

PR #390, opened 2023-12-28
Fix compilation on musl by implementing a custom basename

### LINKED ISSUE #389: Build error using clang 17 on musl (call to undeclared function 'basename')
Hello!
Musl has removed the declaration of `basename` from `string.h`, which now causes a build error when using clang 17+ on musl.
https://git.musl-libc.org/cgit/musl/commit/?id=725e17ed6dff4d0cd22487bb64470881e86a92e7

Error log snippet:
```
--- finit-mdadm.o ---
mdadm.c:61:12: error: call to undeclared function 'basename'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
   61 |                         array = basename(gl->gl_pathv[i]);
      |                                 ^
mdadm.c:61:10: error: incompatible integer to pointer conversion assigning to 'char *' from 'int' [-Wint-conversion]
   61 |                         array = basename(gl->gl_pathv[i]);
      |                               ^ ~~~~~~~~~~~~~~~~~~~~~~~~~
2 errors generated.
*** [finit-mdadm.o] Error code 1

make[2]: stopped in /build/iglunix/init/finit/src/finit-4.6/src
```
Full build log: https://pastebin.com/zQqms8X7

A possible solution would be to include `libgen.h` in the afflicting .c files to use the posix implementation of `basename`.
I do not know if there is a better solution to fix the issue.

Stargirl-chan:
This PR fixes #389.
Tested on a glibc and musl system.

### Changes
- Implemented a custom `basename()` version: `basenm()` using the simple implementation of it that was present in `maybe_fopen_alias()`
- Modified the simple implementation of it to return a copy of the argument that was given, instead of `NULL` if no `/` was present
- Changed all calls to `basename()` to our custom version
- Changed the comparison in `maybe_fopen_alias()` as we no longer return `NULL`

### DIFF
--- a/src/plugin.c
+++ b/src/plugin.c
@@ -28,6 +28,7 @@
 #include <dirent.h>		/* readdir() et al */
 #include <poll.h>
 #include <string.h>
+#include <libgen.h>
 #ifdef _LIBITE_LITE
 # include <libite/lite.h>
 # include <libite/queue.h>	/* BSD sys/queue.h API */
@@ -86,9 +87,11 @@ int plugin_register(plugin_t *plugin)
 	if (!plugin->name) {
 #ifndef ENABLE_STATIC
 		Dl_info info;
+		char *dli_fname = strdup(info.dli_fname);
 
-		if (dladdr(plugin, &info) && info.dli_fname)
-			plugin->name = basename(info.dli_fname);
+		if (dladdr(plugin, &info) && dli_fname)
+			plugin->name = basename(dli_fname);
+		free(dli_fname);
 #endif
 		if (!plugin->name)
 			plugin->name = "unknown";
--- a/src/plugin.c
+++ b/src/plugin.c
@@ -28,7 +28,6 @@
 #include <dirent.h>		/* readdir() et al */
 #include <poll.h>
 #include <string.h>
-#include <libgen.h>
 #ifdef _LIBITE_LITE
 # include <libite/lite.h>
 # include <libite/queue.h>	/* BSD sys/queue.h API */
@@ -44,6 +43,7 @@
 #include "private.h"
 #include "service.h"
 #include "sig.h"
+#include "util.h"
 
 #define is_io_plugin(p) ((p)->io.cb && (p)->io.fd > 0)
 #define SEARCH_PLUGIN(str)						\
@@ -87,11 +87,9 @@ int plugin_register(plugin_t *plugin)
 	if (!plugin->name) {
 #ifndef ENABLE_STATIC
 		Dl_info info;
-		char *dli_fname = strdup(info.dli_fname);
 
-		if (dladdr(plugin, &info) && dli_fname)
-			plugin->name = basename(dli_fname);
-		free(dli_fname);
+		if (dladdr(plugin, &info) && info.dli_fname)
+			plugin->name = basenm(info.dli_fname);
 #endif
 		if (!plugin->name)
 			plugin->name = "unknown";
@@ -256,7 +254,7 @@ void plugin_run_hook(hook_point_t no, void *arg)
 
 	PLUGIN_ITERATOR(p, tmp) {
 		if (p->hook[no].cb) {
-			dbg("Calling %s hook n:o %d (arg: %p) ...", basename(p->name), no, arg ?: "NIL");
+			dbg("Calling %s hook n:o %d (arg: %p) ...", basenm(p->name), no, arg ?: "NIL");
 			p->hook[no].cb(arg ? arg : p->hook[no].arg);
 		}
 	}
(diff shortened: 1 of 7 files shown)


src/plugin.c
```diff
@@ -86,9 +87,11 @@ int plugin_register(plugin_t *plugin)
 	if (!plugin->name) {
 #ifndef ENABLE_STATIC
 		Dl_info info;
+		char *dli_fname = strdup(info.dli_fname);
 
-		if (dladdr(plugin, &info) && info.dli_fname)
-			plugin->name = basename(info.dli_fname);
+		if (dladdr(plugin, &info) && dli_fname)
+			plugin->name = basename(dli_fname);
+		free(dli_fname);
```
troglobit, line 94: Why this code?  There is no mention of it in the PR, the commit message, or the original issue filing explaining why you believe it's needed.

### Code review

 1. You do not check the return value from `strdup()`, it may be `NULL`
 2. You assign `plugin->name` a subset of the allocated memory and then `free()` it?  That means `plugin->name` will point to a region of memory that may be reused soon for other things.
 3. You don't need to `strdup()`, the `basename()` function finds the last `/` (if any) and returns a pointer the the char right after it.  No other use of `basename()` in the code uses a `strdup()`.

### Final comment

Also, you missed one use of `basename()` in `plugin/modprobe.c`.  Not sure if one of the other includes in turn include `libgen.h`, or if your build does not have the modprobe plugin enabled.

troglobit (changes requested):
Please remove the extra code (`strdup()`) and add an include also in `plugin/modprobe.c`.  Thank you for taking care of this!

src/plugin.c
Stargirl-chan: Oh apologies for forgetting to mention it anywhere.
Why is that here:
- posix basename expects to be given a `char*` instead of a `const char*`, in fact the [spec](https://pubs.opengroup.org/onlinepubs/007904975/functions/basename.html) and the man page for [basename(3)](https://man7.org/linux/man-pages/man3/basename.3.html) mention, that the given string may be modified. This would result in undefined behavior if it tries to modify a read-only memory address.
> The basename() function may modify the string pointed to by path, and may return a pointer to static storage that may then be overwritten by a subsequent call to basename().

> Both dirname() and basename() may modify the contents of path, so
       it may be desirable to pass a copy when calling one of these
       functions.
       
- The goal was to create a copy of `info.dli_fname` that is allowed to be modified to adhere to the spec

### Review answers
1. Good catch, it went over my head that `strdup()` may return `NULL`
2. Yes the `free()` makes absolutely no sense there
3. Others uses of `basename()` in the code base do not require the usage of `strdup()`, because they are all a char pointer, opposed to this specific instance, where it is a `const char*`

### Final comment answers
`plugin/modprobe.c` does not have a call to `basename()`? It has a local variable named `basename` in the `maybe_fopen_alias()` function.
But I forgot one in `tmpfiles.c`

troglobit: OK, this is a mess. I've just had a look at the source implementation of `basename()` in both musl and glibc.  The xpg versions do indeed modify the contents, even though there's no need for it.  The GNU version of `basename()` in GLIBC, however, does not. So I guess that's what I looked the last time I dug into this.

I think for consistency, and to keep things sane, we need to carry our own version of `basename()`, e.g., `basenm()` that basically does what parts of `maybe_fopen_alias()` does.  That would be my solution.

Stargirl-chan: I think that is the best solution going forth. Would a suitable place for the custom `basenm()` function be `src/helpers.c`?

troglobit: Put it with `progname()` and `str()` in `src/util.c`, we may very possibly want to reuse this in `initctl` and other programs later.  Thanks!

src/util.c
```diff
@@ -151,6 +151,24 @@ char *progname(char *arg0)
...
+
+	basename = rindex(path, '/');
+	if (!basename)
+		return (char *) path;
+
+	basename++;
+	return basename;
+}
+
 char *str(char *fmt, ...)
```
troglobit, line 172: The comment is very confusing, the function never returns a copy.  Just simplify it as:

```C
/* basename(3) replacement that does not modify its argument */
```

Please follow the coding style of the project, or at least, surrounding code.  The opening brace of the function goes on the next line.

Also, I'm not overly happy with the conversion from const to non-const here.  But making the function return a `const char *` probably causes a ripple effect on more of the code?

troglobit (changes requested):
Other than the minor comments I've made (below), this looks good!

src/util.c
Stargirl-chan: Will push the changes in a few seconds.

Yes, sadly having it return `const char *` will cause a ripple effect, though currently only 2 places in the code base come to mind (as of now), `src/mdadm.c:62` and `src/plugin.c:92` where the return value of `basenm()` is assigned to `char *` in both cases.

`src/mdadm.c:62` could be made `const char *` without problem
`src/plugin.c:92` not so much, as it's reassigned and from the looks of it, not so trivial to make it `const char *` as well

troglobit: OK. Makes sense. I'll probably have a look at it myself later on. 

Thank you for taking the time to report and fix this!

troglobit (approved):

PR #393, opened 2024-01-09
Improve wording.

solt87:


README.md
```diff
@@ -397,15 +397,15 @@ in [doc/plugins.md](doc/plugins.md#hooks).
 ### Limitations
 
 It is not possible to call Finit via signals or use `initctl` in any
-runparts or `/etc/rc.local` script.  This because Finit is single
```
troglobit, line 400: I believe this is correct English.

README.md
```diff
@@ -397,15 +397,15 @@ in [doc/plugins.md](doc/plugins.md#hooks).
...
+runparts or `/etc/rc.local` script.  This is because Finit is single
 threaded and is calling these scripts in a blocking fashion at the end
 of runlevel S, at which point the event loop has not yet been started.
 
 The event loop is the whole thing which Finit is built around, except
 for runlevel S, which remains a slow procession through a lot of set up,
 with a few hooks and blocking call outs to external scripts.
 
-However, not all `initctl` commands are prohibited, supported commands:
+However, not all `initctl` commands are prohibited, supported commands
```
troglobit, line 408: I'd rather drop 'are:'

troglobit (changes requested):

troglobit:
ping?

README.md
solt87: I'm not a native speaker, but I still think a verb phrase is missing. "This is" would be the main clause, "because Finit is [...]" would be it's dependent clause.

README.md
solt87: If we drop the last "are", then it would be better to make "Supported commands:" into its own sentence. In it current form, the two parts seem to be coordinated clauses, in which second one needs a verb phrase just as the first one. (But I wouldn't bet on it, then again, it just feels odd.)

troglobit: Yes, good point.  Please change to two separate sentences.

README.md
troglobit: I must persist, I am quite sure this is correct English.  Please drop this change.

troglobit:
Great work, could you please also squash the two commits to one?

solt87:
I'm in the process of searching for a way to squash my commits in the web UI. (I had hard time finding how to edit(?) my commit in the first place.) Nothing useful so far, although they say you have the option to squash-merge a PR. (Sorry for the delay. Usually I use git on the command line, I don't think I performed anything more complicated than a simple unedited commit+PR in the web interface...)

troglobit (approved):

PR #406, opened 2024-05-09
configure.ac: make cgroup2 configurable

liuming50:
We would like cgroup2 to be configurable at build time, when it's disabled, finit will not check /sys/fs/cgroup for cgroup2 support.

The reason for why we need this is because we'd like to use our own rules for controlling cgroup2, and it's hard to adapt to finit's default settings.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -395,7 +395,7 @@ Optional features:
   Built-in sulogin......: $with_sulogin $sulogin
   Built-in watchdogd....: $with_watchdog $watchdog
   Built-in logrotate....: $enable_logrotate
-  Check cgroup2.........: $enable_cgroup2
+  Use cgroup v2.........: $enable_cgroup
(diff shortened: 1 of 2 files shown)


troglobit:
Looks OK overall, but I don't want to break the default, so please flip the setting to --disable-cgroups so it doesn't hit users upgrading.

liuming50:
@troglobit thanks for the review! Have changed it to "--disable-cgroup2" in V2 patch set.

troglobit (approved):
Made some minor changes: ordering and simplify naming + help text.  Hope that's OK with you?

liuming50:
@troglobit  > Made some minor changes: ordering and simplify naming + help text. Hope that's OK with you?

Sure, thanks for the modification!

troglobit:
Alright, thanks for taking the time to improve Finit! 😃🙏

PR #408, opened 2024-08-05
Avoid remounting already mounted /run and /tmp directories

mandelmassa:
This is a patch for finit 4.8 beta 1.

It adds the function fistmpfs to determine if a new tmpfs mount should be performed on /run and /tmp. The function supports cases where more complex mount hierarchies are in use, including overlayfs backed mounts.

In our system, /tmp is a symbolic link to a subdirectory in an overlayfs mount. This means that /tmp does not explicitly show up in /proc/mounts, so in fs_finalize, finit decides to mount a new tmpfs on top of it.

This turns out to be problematic for our smallest systems where the new tmpfs on /tmp is not large enough.

I do realize this may be a bit particular to our specific setup, but the patch works on our system. Any comments are welcome.

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -42,6 +42,8 @@
 # include <sys/ioctl.h>
 #endif
 #include <sys/sysinfo.h>	/* sysinfo() */
+#include <sys/vfs.h> /* statfs */
+#include <linux/magic.h>
 #ifdef _LIBITE_LITE
 # include <libite/lite.h>
 #else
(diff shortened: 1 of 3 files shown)


src/util.c
```diff
@@ -572,6 +574,18 @@ int fismnt(char *dir)
...
 }
 
+/* Return 1 if dir is a backed by tmpfs or overlayfs */
+int fistmpfs(char *dir)
+{
+	struct statfs info;
+
+	statfs(dir, &info);
+	if (info.f_type == TMPFS_MAGIC || info.f_type == OVERLAYFS_SUPER_MAGIC)
+		return 1;
```
troglobit, line 584: Either initialize the stack variable with `{ 0 }`, and/or check the return value of `statfs()`, e.g., on error `return 0;`

troglobit (changes requested):
Please fix the review comment, otherwise this looks OK to me :+1:

troglobit (approved):

PR #419, opened 2024-11-03
RTC plugin fixes and extension

troglobit:
This PR contains three major changes to the RTC plugin:

 - Big refactor for readability, breaking out `time_set()` and `time_get()` functions
 - Add support for saving and restoring time from a file on systems with seriously buggy RTC
 - Fix `--with-rtc-date` fixes, both configure script and `strpdate()` error handling


troglobit:
@wkz could you have a look at the second commit in this PR?  It should fix issue kernelkit/infix#794 -- I have tested it on an `x86_64` qemu system where I disabled the RTC driver, and it works great.  We should test it also on aarch64, of course, but it needs a few tweaks to the finit.mk in Infix first.

wkz (approved):
This looks great! :muscle: 

### Potential Improvement
We could opt to _always_ keep an RTC file when that has been enabled in the config - i.e., even when a real RTC is available. 

If the time we read from the RTC is invalid, then we could use the file data as a fallback. This would improve the worst-case date offset for a device that has been gathering dust for a year, and whose super-cap has long since been drained.

In combination `--with-rtc-date`, that should ensure that the device always comes up with a date good enough to take care of most certificate ranges.

troglobit:
> This looks great! 💪

Thanks! :smiley: 

> If the time we read from the RTC is invalid, then we could use the file data as a fallback. This would improve the worst-case date offset for a device that has been gathering dust for a year, and whose super-cap has long since been drained.

Interesting idea!  And, like you say, combined `--with-rtc-date`, it should improve the situation radically!  I'll have a look!

troglobit:
Alright, now it looks as follows if both RTC is bad and the system has no "save game" on disk yet:

```
● ● ●  Infix -- a Network Operating System v24.10.1-68-ge50fa66b ═══════
[ OK ] Mounting filesystems from /etc/fstab
[WARN] Restoring system clock (UTC) from RTC
[WARN] Restoring system clock from backup
```

PR #428, opened 2025-02-19
allow sysctl path to be configured at build time

aanderse:
my `sysctl` isn't located in `/sbin` - i followed convention seen elsewhere

### DIFF
--- a/plugins/procps.c
+++ b/plugins/procps.c
@@ -32,6 +32,10 @@
 #include "helpers.h"
 #include "plugin.h"
 
+#ifndef SYSCTL_PATH
+#define SYSCTL_PATH "/sbin/sysctl"
+#endif
+
 static void setup(void *arg)
 {
 	glob_t gl;
@@ -54,7 +58,7 @@ static void setup(void *arg)
 		for (i = 0; i < gl.gl_pathc; i++) {
 			char cmd[160];
 
-			snprintf(cmd, sizeof(cmd), "/sbin/sysctl -e -p %s >/dev/null", gl.gl_pathv[i]);
+			snprintf(cmd, sizeof(cmd), "%s -e -p %s >/dev/null", SYSCTL_PATH, gl.gl_pathv[i]);
 			run(cmd, "sysctl");
 		}
 		globfree(&gl);


troglobit (changes requested):
Great idea to make this configurable!  You'll undoubtedly find more.

The format of these have been a bit up in the air.  For internal ones, and those sourced from paths.h for instance, they are usually prefixed with a `_PATH`, but for those read at `configure` time I think we should go with `FOO_PATH`.

Example (which needs updating): `modules-load.c` use `MODPROBE_PATH`, which needs an `#ifndef` like yours in `procps.c`. 

So please go for `SYSCTL_PATH` instead, and if you'd like to update more plugins to follow, you're very welcome to do so :smiley: :+1:

aanderse:
> So please go for `SYSCTL_PATH` instead

ah, yeah i see now - thanks

> You'll undoubtedly find more.

well... most of the others are different - they have the implicit assumption that `PATH` contains the executable in question, which is probably the best way to do it

`procps.c` is different than most other cases because it hardcodes a path (`/sbin`)
even your `modules-load.c` example isn't quite the same because it allows you to specify a path at "runtime" (ie. in the configuration file)

in fact... maybe we should just modify `procps.c` to assume `sysctl` exists on `PATH` like the rest... or are there implications to that i haven't considered? :thinking:

troglobit:
> > You'll undoubtedly find more.
> well... most of the others are different - they have the implicit assumption that `PATH` contains the executable in question, which is probably the best way to do it

Hmm, yeah :thinking: 

> `procps.c` is different than most other cases because it hardcodes a path (`/sbin`) even your `modules-load.c` example isn't quite the same because it allows you to specify a path at "runtime" (ie. in the configuration file)
> in fact... maybe we should just modify `procps.c` to assume `sysctl` exists on `PATH` like the rest... or are there implications to that i haven't considered? 🤔

For most users that would probably be sufficient, yes.  But I like that with your PR we now will get a way for users to override.

Also, as PID 1 the default `PATH` is usually `_PATH_STDPATH` which, unless user override, defaults to `"/usr/bin:/bin:/usr/sbin:/sbin"`

aanderse:
out of curiosity what is the difference between `sysctl` and ` mount` in your mind then? in `finit.c` there is no way to satisfy requirements other than ensuring `mount` exists on `PATH` - no `MOUNT_PATH` (or i guess `_PATH_MOUNT` 😅) exists - while `procps.c` will ignore `PATH` and only observe compile time options. 

maybe i am thinking too hard about a fairly small detail...

aanderse:
great answer, what i was looking for

in that case... how about we merge this as it stands and in the future if i want i can reference this in a new PR?

thanks

troglobit:
Yup 💯 just wanted to sync with you first.

troglobit (approved):

PR #439, opened 2025-05-24
tmpfiles: refactor into separate executable

### LINKED ISSUE #423: tmpfiles: initctl command or native cleanup-age time
i took a quick look and it doesn't seem like i can explicitly call tmpfiles functionality - no separate binary or invocation like `initctl tmpfiles --clean/--create` or whatever

i also didn't see any documentation for how often/when cleanup and creation functionality runs automatically...

maybe i missed documentation? and would it be possible to explicitly run tmpfiles, either as a separate executable or an initctl sub command?

thank you for your time and consideration 🙇‍♂

aanderse:
this PR is an attempt to resolve #423

i thought i would open this PR early for more concrete discussion and feedback - i don't want to end up too far down a path which you aren't interested in merging

i am proposing the following for this PR:
- [x] simply split tmpfiles into its own binary under `libexec/`
- [x] add `--create` and `--remove` args to the program so we can run only the desired behaviour (`r` and `R` rules will execute when `--remove` is passed to the program, the rest when `--create` is)

this sets us up for future PR(s) where more _useful_ functionality can be fleshed out, including the `--clean` flag for time based cleanup

---

> [@troglobit](https://github.com/troglobit) 
> Yeah a `tmpfilesd` or similar would be neat for this.  Then we could relocate all the current code to that, and call it with an `--init` flag or something at bootstrap, before we start it as a proper service.

 _Originally posted by @troglobit in [#423](https://github.com/troglobit/finit/issues/423#issuecomment-2900117750)_

regarding `tmpfiles` as a service... i'm not sure how useful that actually is - the main reason new tmpfile configs are added to the system is part of package/system management hooks - but of course there is still the cleanup aspect... given that there is no cleanup currently implemented, i don't think that is a strong argument for a service

what do you think? am i heading in a direction you would be willing to merge? thanks again for your time and consideration :bowing_man:


src/tmpfiles.c
```diff
@@ -23,12 +23,99 @@
...
+#else
+	struct group *grp;
+
+	if ((grp = getgrnam(group)) == NULL)
+		return -1;
+
+	return grp->gr_gid;
+#endif
+}
+
```
aanderse, line 118: this is now duplicated code from the core of `finit` - should it be pulled into separate `.c/h` files for proper code reuse?

troglobit: Yeah it definitely should, no code duplication please :smiley:

src/tmpfiles.c
```diff
@@ -449,7 +536,7 @@ static void tmpfiles(char *line)
 
 		for (size_t i = 0; i < gl.gl_pathc; i++) {
 			snprintf(buf, sizeof(buf), "restorecon %s %s", opts, gl.gl_pathv[i]);
-			run(buf, "restorecon");
+			// run(buf, "restorecon");
```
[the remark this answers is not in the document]
troglobit: Parameter or #ifdef if built with Finit proper, I leave that up to you to decide. Either works :+1:

troglobit (commented):
Looks good so far, nice work! :+1: 

Please elaborate a little more on command line options though. You mention `--create`, which I was sort of expecting in the call in `bootmisc.c`.  A basic usage text would be sufficient.

src/tmpfiles.c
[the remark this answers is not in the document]
aanderse: i wasn't sure what you meant here...

troglobit:
> * i would like to squash the second commit into the first, but i think you probably have strong opinions about the second commit... so i'll let you comment on that first

You're right, I have a lot of opinions about that commit :smile: I'll try to be harsh but fair:

 1. I do not like moving out all that code from `exec.c`.  Related code should go in the same file, or it will be very difficult to keep track of where it all is.
 2. Why do we have to create `misc.c`?  Why can't the new application link against the same source files as Finit?  Sure not all functions are used, but dead code is usually something the linker can clean up.
 3. We already have `util.c`, `helpers.c`, giving it the name `misc.c` just adds to a mess that might need to be untangled in another way.  In the example of of moving out `sig_unblock()` (which I don't like either) one could argue that generic signal code that can be shared should go in a `sig-common.c` instead.  Similarly for the functions from `exec.c`, could be moved to a `exec-common.c`

In summary, please explain to me why `tmpfiles_SOURCES` cannot include `exec.c helpers.c sig.c util.c` instead.

Also, your commit messages need to tell the story of *Why* a change is made, reasoning with a future reader, explaining the intent.

>     * i noticed there was code duplication in `src/runparts.c`... at least one function there now exists in the new `src/misc.c`, so that could be removed in a future PR

OK, nice find.

>     * i haven't booted into this to test it yet

For future reference, and as a matter of process, please keep PRs as *draft* until you've verified your changes.

> hopefully this is in the right direction... looking forward to feedback

Definitely in the right direction, great work so far!

aanderse:
> 1. I do not like moving out all that code from `exec.c`.  Related code should go in the same file, or it will be very difficult to keep track of where it all is.
> 
>     2. Why do we have to create `misc.c`?  Why can't the new application link against the same source files as Finit?  Sure not all functions are used, but dead code is usually something the linker can clean up.
> 
>     3. We already have `util.c`, `helpers.c`, giving it the name `misc.c` just adds to a mess that might need to be untangled in another way.  In the example of of moving out `sig_unblock()` (which I don't like either) one could argue that generic signal code that can be shared should go in a `sig-common.c` instead.  Similarly for the functions from `exec.c`, could be moved to a `exec-common.c`
> 
> 
> In summary, please explain to me why `tmpfiles_SOURCES` cannot include `exec.c helpers.c sig.c util.c` instead.

this was not my original intention - i also wanted to link existing `.c` files. i guess i misunderstood your prior comments because this seemed off to me too. and you maybe you misunderstood the problems i was having. i will come up with a summary of the problems i had and be very explicit because we aren't understanding each other properly.


> Definitely in the right direction, great work so far!

thanks for the feedback - more to come

src/tmpfiles.c
```diff
@@ -23,12 +23,53 @@
...
+char *sdown;
+char *hostname;
+int bootstrap;
+int single;
+int readiness;
+char *finit_conf;
+char *fstab;
+int prevlevel;
+int cmdlevel;
+int kerndebug;
```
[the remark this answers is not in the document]
troglobit: No worries (sorry for the late reply)!  Yeah, you've found something really stinky that comes from the legacy of being single-target (only finit binary) for so long.

I've been meaning to either a) at least move each of these to their owner C file, or b) create a `struct global object` or similar.

In the case of (a), taking `cfglevel` as a shining example, it should live privately in and be owned by `src/conf.c`, and there should be setter/getter methods to access it.  E.g., from `src/api.c` it would call `conf_set_cfglevel(lvl)`.  

Sorry for the mess!

aanderse:
@troglobit i loaded this branch of `finit` onto my system last night - looks like it is working as expected:
- i saw the expected message at boot about creating directories and files
- a few test rules i added before reboot were run as expected
- manually tested a few different types and executed `--create`, `--remove`, as well as `--create --remove` together

when you get a chance please let me know how this looks... thank you!

troglobit (changes requested):
Looks much better, nice work!

You can squash the first two commits, but the last one needs to be split up so you do any refactoring/reshuffle of code in a separate commit from where you add --create and --remove flags.  Also, the first line of that commit wraps, try to keep a brief first line of the commit, e.g., "tmpfiles: add --create and --remove flags" is sufficient, the rest of the text can go in the body.

aanderse:
@troglobit i think the second commit accomplishes what you want to see here

don't hesitate to let me know if there are _any_ changes i can make so you're completely happy with this PR... your code quality is great, so i really don't mind nitpicks from you :bowing_man:

src/tmpfiles.c
```diff
@@ -279,182 +319,223 @@ static void tmpfiles(char *line)
...
+			if (!strc) {
+				struct dirent **namelist;
+				int num;
+
+				if (!S_ISDIR(st.st_mode))
+					break;
+				num = scandir(path, &namelist, NULL, NULL);
+				free(namelist);
+				if (num >= 3)
+					break; /* not empty */
```
troglobit, line 406: Here we will (potentially) leak memory.  You need to iterate over all the num, when num > 0, freeing all the entries in namelist[i], also handling the case when num == -1.

I suggest splitting out this code to a helper function, e.g., `is_dir_empty()` or similar.

src/tmpfiles.c
```diff
@@ -279,182 +319,223 @@ static void tmpfiles(char *line)
...
+			}
+			if (fisdir(arg) && !fisslashdir(arg)) {
+				size_t len = strlen(arg) + 2;
+
+				dst = malloc(len);
+				if (!dst)
+					break;
+				snprintf(dst, len, "%s/", arg);
+				arg = dst;
```
troglobit, line 415: Another memory leak.  Consider using `alloca()` instead, similar to your using `strdupa()` below.

troglobit (changes requested):
This looks *much* better, very nice!

My comments on memory usage could be considered theoretical, but I've learned that it's always best to keep it clean because you never know how code transforms or where it is used.  An esoteric use-case is running Finit on uClinux systems (CPUs without MMU), there Linux would *not* clean up and free memory a process has not freed itself.

src/tmpfiles.c
aanderse: is it? `dst` is `free`d at the end the function with no other exit point before that takes place... :thinking:

src/tmpfiles.c
aanderse: great idea! see additional commit i have pushed

troglobit:
:1st_place_medal: Great work on this, time to squash some of the commits and mark this as *Ready for review* :checkered_flag:

troglobit:
> awesome! thanks for all your time and patience on this!

No problem at all!

> previously you mentioned that you like commits to be well defined logically small changes, which sounds good - i feel that's what we have here... so please let me know explicitly what you want and i am happy to comply

Yeah, so logical commits are just that: "whitespace changes only", "drop foo functionality, not used anymore", "add bar to baz, extending xyzzy", just cleanly separating what removes, relocates, and what adds new functionality.  Think; "what if I had to debug these changes myself", what would you like to see.  Also, there's the art of bisectability, i.e., each change builds and is testable on its own.  This latter one I struggle with myself.

The most important thing, apart from commit messages telling *why* a change is made, is to keep changes logical.  So, when you look at a series of patches/commits you should be able to, at some degree, trust the submitter.  So when they say, "whitespace changes only", you can skip that commit at the beginning and focus on removals and additions.  It's all about making it easier for the reviewer, and yourself a few weeks/months/years down the line ...

aanderse:
i have been running this for a bit on my system and i think it is ready for a final review

troglobit (approved):
This looks great, and the tests pass, so time to merge!

Awesome job on this, @aanderse ! :100: :1st_place_medal:

aanderse:
thanks again!

PR #458, opened 2025-11-28
Add support for Linux capabilities

### LINKED ISSUE #454: Capabilities support
a pretty common pattern for many daemons is to require a very minimal set of root privileges, like the ability to bind to a port under 1024, and no other reason - web servers, dns servers, etc... for example

i think that capabilities support in `finit` would be a nice feature

the [dinit](https://github.com/davmac314/dinit) service manager, which has some similar goals to `finit`, has this support - which i think is supporting evidence that capabilities support wouldn't be project scope creep

aanderse:
Implement Linux capability support for services, allowing them to run with minimal required privileges instead of running as root. This uses the modern IAB (Inheritable, Ambient, Bounding) API from libcap.

---

- i am still working on the documentation but i thought it was worth pushing this as-is in case i'm going in the wrong direction
- testing some simple scenarios on my system is fine so far :crossed_fingers:

resolves #454

---

my motivation for this feature: i recently installed `finit` on my home server which runs `caddy` as a web server `cap_net_bind_service` would be very helpful here

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,7 +34,7 @@ jobs:
       - name: Install dependencies
         run: |
           sudo apt-get -y update
-          sudo apt-get -y install pkg-config tree jq libuev-dev libite-dev
+          sudo apt-get -y install pkg-config tree jq libuev-dev libite-dev libcap-dev
       - uses: actions/checkout@v4
       - name: Static Finit
         run: |
--- a/configure.ac
+++ b/configure.ac
@@ -72,6 +72,10 @@ AC_ARG_ENABLE(cgroup,
         AS_HELP_STRING([--disable-cgroup], [Disable cgroup v2 support, default: autodetect from /sys/fs/cgroup]),,[
         enable_cgroup=yes])
 
+AC_ARG_ENABLE(libcap,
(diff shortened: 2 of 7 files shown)


troglobit (changes requested):
Minor comment, so far this looks great! 💯 Really looking forward to reading the documentation so I might actually learn capabilities, finally 😅

aanderse:
i tested some services and everything worked as expected

@troglobit is there anything else you would like me to add or change before i mark this as ready for review?

troglobit:
> i tested some services and everything worked as expected

Cool! 

> @troglobit is there anything else you would like me to add or change before i mark this as ready for review?

Nope, if you're happy and feel ready then please go ahead 😊

aanderse:
i'm so sorry for the hassle @troglobit - i am very confused how none of my daemons running as `root` had any trouble when running this code... then again, because of this code i was able to run most daemons as a non `root` user... :sweat_smile:

the documentation and code changes make sense to me. i will give these changes a test today and report back. thank you!

troglobit:
> i tested this on my home server which makes use of `cap_net_bind_service` on several services and found no regressions ✔️

Awesome, that's great news!

> thanks again @troglobit 🙇‍♂️

Thank you for testing! :pray:

PR #461, opened 2025-12-19
Add support for supplementary groups

aanderse:
Implement supplementary group support for services, allowing them to access resources owned by multiple groups. Uses the @user:group,sup1,sup2 syntax to explicitly specify supplementary groups, as finit does not read group membership from /etc/group.


aanderse:
amazing! thanks for the review @troglobit :bowing_man:

troglobit:
Sorry for the delay, I'm all caught up here now.  It [seems to me](https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#SupplementaryGroups=) systemd defines this as:

> "this option does not override, but extends the list of supplementary groups configured in the system group database for the user"

Which I interpret as we should read and apply both the system groups for the user and this additional set provided in this PR, if `@user` is provided.

(I wasn't even aware of [setgroups(2)](https://linux.die.net/man/2/setgroups) until this PR.)

troglobit:
> > Which I interpret as we should read and apply both the system groups for the user and this additional set provided in this PR, if `@user` is provided.
> sure, we can match `systemd` exactly in behaviour then, i'll look into this at some point and update the PR :+1:

Thanks mate, really appreciate!

aanderse:
ok... ran a few tests on my system and validated the new behaviour - should be good to go :+1:

troglobit (approved):
LGTM. Nice work on this!

PR #463, opened 2025-12-31
Add switch_root support for initramfs to real root transitions

aanderse:
Implement switch_root functionality allowing Finit to serve as the init in an initramfs, then transition to the real root filesystem.  Useful for systems requiring early boot tasks like LUKS unlock, LVM activation, or network boot before mounting the real root.

Adds INIT_CMD_SWITCH_ROOT API command, `initctl switch-root` subcommand, and HOOK_SWITCH_ROOT plugin hook point.  The implementation gracefully stops services, moves virtual filesystems (/dev, /proc, /sys, /run) to the new root, deletes initramfs contents to free memory, then execs the new init as PID 1.

See GitHub Discussion #292 for background.

---

this is what my system is running on now :rocket:

milestone `4.16`?


doc/switchroot.md
```diff
@@ -0,0 +1,149 @@
...
+1. Re-read `/etc/finit.conf` from the new root
+2. Re-initialize all conditions
+3. Start services according to the new configuration
+
+
+See Also
+--------
+
+- `switch_root(8)` - BusyBox switch_root utility
+- `initramfs(7)` - Linux initramfs documentation
```
troglobit: Please create actual links to man pages, similar to what I did for `capabilities.md`.  I used https://man7.org which seems like a good enough provider.

troglobit (changes requested):
Wow, amazing work on this! 💪💯 

I have a few comments and suggestions, as well as a question: do you think this would be worth having a regression test for?  Not sure if it fits with the current test framework, might actually need a separate repo even that triggers an entire circus of events ...

aanderse:
thanks for the thorough review @troglobit! hopefully this covers it... note that i did add one value to `src/private.h` which may not be desirable...? let me know if not!

> do you think this would be worth having a regression test for? Not sure if it fits with the current test framework

currently i'm working on [this](https://github.com/finix-community/finix/blob/82d18b4af05a875dd6adb1f9e8ac3d8c3c88bccc/tests/finix-test-driver/shell.nix) as an example integration test - it allows you to define and build a `finix` machine (kernel, initramfs, etc...) for `qemu`, launches it, then allows you to interact via a `testScript` attribute (`tcl` script) with the vm (or vms)

[here](https://asciinema.org/a/E9cDg66oRZ81ljFabMpCOrLN7) is an example output run from the above

maybe this seems like a big black box... but if you think this has any value to `finit` i would love to discuss with you and see if i can provide any value for you... end to end os integration tests are a very strong point for nixos (and finix now too)

troglobit:
LGTM and the tests pass, let's merge! :sunglasses:

PR #470, opened 2026-01-19
Create mount points in fs_init() if they don't exist

aanderse:
In containerized or virtualized environments, standard mount point directories may not exist at boot.  Ensure they are created before attempting to mount.

---

when running `finix` in a container we can skip stage 1 (initramfs) entirely - this change let's me boot _directly_ into `finit` in my stage 2 without the need for a shell script to `mkdir` and then call `finit`! :rocket:

### DIFF
--- a/src/finit.c
+++ b/src/finit.c
@@ -523,6 +523,10 @@ static void fs_init(void)
 		if (fismnt(fs[i].file))
 			continue;
 
+		/* Create mount point if it doesn't exist */
+		if (mkdir(fs[i].file, 0755) && errno != EEXIST)
+			warn("Failed creating mountpoint %s: %s", fs[i].file, strerror(errno));
+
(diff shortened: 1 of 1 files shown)


src/finit.c
```diff
@@ -523,6 +523,10 @@ static void fs_init(void)
 		if (fismnt(fs[i].file))
 			continue;
 
+		/* Create mount point if it doesn't exist */
+		if (!fisdir(fs[i].file))
+			mkdir(fs[i].file, 0755);
```
troglobit: I mostly run Finit on systems with read-only root (squashfs), so I'm used to it the hard way, but this is better, much more user-friendly 😃 

What do you think about skipping `fisdir()` and just:

```c
if (mkdir(...) && errno != EEXIST)
    warn("Failed creating mountpoint %s", ...); // strerror(errno) is appended
```

This may bombard the user with a lot of warnings, but at bringup you want to know, right?

PR #483, opened 2026-03-17
Add Plymouth boot splash plugin

aanderse:
Manage the plymouthd lifecycle across boot, switch_root, and shutdown. Activated by the "splash" kernel command line argument.

---

see it in action [here](https://github.com/finix-community/finix/discussions/1#discussioncomment-16184663)

### DIFF
--- a/plugins/plymouth.c
+++ b/plugins/plymouth.c
@@ -1,0 +1,293 @@
+/* Plymouth boot splash plugin for Finit
+ *
+ * Copyright (c) 2012-2026  Aaron Andersen <troglobit@gmail.com>
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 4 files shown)


plugins/plymouth.c
```diff
@@ -0,0 +1,310 @@
...
+	char cmd[256];
+	int rc;
+
+	if (plymouth_alive())
+		return;
+
+	/* Mount devpts if not already mounted (needed for --attach-to-session) */
+	if (!fismnt("/dev/pts")) {
+		mkdir("/dev/pts", 0755);
+		mount("devpts", "/dev/pts", "devpts",
```
aanderse: something about this feels not quite right... any opinions or alternatives?

troglobit: Tricky, but maybe we can move the entire block of `fs_mount("devpts" ...` from `fs_finalize()` to `fs_init()`?  That way this code right here wouldn't be needed.

plugins/plymouth.c
```diff
@@ -0,0 +1,310 @@
+/* Plymouth boot splash plugin for Finit
+ *
+ * Copyright (c) 2012-2026  Joachim Wiberg <troglobit@gmail.com>
+ *
```
troglobit: I think you can claim this one for yourself 😃

troglobit (changes requested):
Looks great overall, absolutely LOVE the video 🤩 nice work!

plugins/plymouth.c
aanderse: i tried a new approach based on your comments... mount initially, but then remount so we don't lose/alter the existing behavior. let me know what you think about this approach...

i will test on my machine tonight hopefully

---

as always, thank you so much for your support :bowing_man:

aanderse: tested and this works well on my system :+1:

aanderse:
@troglobit if you like the second commit i can squash and will mark as ready

troglobit:
> @troglobit if you like the second commit i can squash and will mark as ready

LGTM 👍

aanderse:
testing looks good - thanks again @troglobit!

troglobit (approved):

PR #487, opened 2026-04-18
tmpfiles.d: update /var/tmp mode

aanderse:
`/var/tmp` mode should match `/tmp`

it isn't explicitly stated in the FHS standard that mode should be `1777` but the implication is there
additionally most distros set it to this value, as does `systemd`
i recently found that `flatpak` depends on these permissions, along with some other applications

---

@troglobit what do you think? is it reasonable to update this value here?


troglobit:
Nice catch!  And yes, I think it's definitely reasonable. Thanks!

PR #491, opened 2026-05-18
Finit 5: udev/eudev replacement

troglobit:
The changes to Finit's `keventd` on this branch, in conjunction with [libudev-zero](https://github.com/illiliti/libudev-zero), should prove to be a good-enough replacement for many systems.

We might adopt libudev-zero to the project in case that's needed, but hopefully we can bring it some new fresh blood instead :vampire: 

> [!IMPORTANT]
> This branch is still an active work in progress, and it may even be dropped and its feature set be moved to the `next` branch, which is the canonical branch for all Finit 5 work.


aanderse:
@troglobit this is _amazing_! this code looks like it is pretty close to providing a full replacement for `udev`, including `udisks` support! thanks for your work on this :bowing_man:

if you get to a point where you think this is good enough i'm more than happy to test it out, just let me know!

troglobit:
> @troglobit this is _amazing_! this code looks like it is pretty close to providing a full replacement for `udev`, including `udisks` support! thanks for your work on this 🙇‍♂️

Glad to hear you like it! :blush: 

> if you get to a point where you think this is good enough i'm more than happy to test it out, just let me know!

It should be fairly complete, except for libudev, which some packages need. You can use [libudev-zero](https://github.com/illiliti/libudev-zero) as a stand-in though. That's how I've been testing in Infix OS.

I should put something like this in the docs to help:

> > libudev-zero is a daemonless libudev.so.1 replacement.  It reads device properties directly from /sys and /run/udev/data/ without requiring a daemon.  Pairs with finit-keventd, which produce /run/udev/data/ records and rebroadcasts uevents on the netlink group libudev-zero monitors.

I'll probably do some commit housekeeping before I port it to the `next` branch where the D-Bus stuff is.  After that I'll have a stab at either a new file format, or the zones idea I had.

aanderse:
> It reads device properties directly from /sys and /run/udev/data/

actually `libudev-zero` doesn't actually read data from `/run/udev/data`... yet 🤞

there is a [fork](https://codeberg.org/Gardenhouse/libudev-zero/src/commit/8573d647bb44efd3569b6adf782bfb947a2b7124/src/udev_device.c#L646) which added this functionality, but it doesn't exist upstream. i hope upstream would welcome a PR implementing this. obviously this is an important step for us.


> You can use [libudev-zero](https://github.com/illiliti/libudev-zero) as a stand-in though. That's how I've been testing in Infix OS.

i would love to hear more about your user cases and testing! `finix` makes extensive use of `libudev-zero` and it works really well for us 👍

aanderse:
i don't have the `udev` binaries like `fido_id` in my `PATH` - this seems to result in a bunch of zombie processes. i don't specifically need these binaries, so would it be fine for `keventd` to gracefully ignore when a `RUN` program doesn't exist?

image

aanderse:
i booted into a graphical environment with `keventd`! `greetd`, my graphical login manager, initially failed because `/dev/dri/...` didn't exist, but then restarted and was fine. `udev` used to deal with this problem via its `settle` task, but that is discouraged these days - applications are expected to have `udev` support and subscribe... or something, i guess :slightly_smiling_face:

do you have any suggestions for `keventd`? `finit` isn't defacto, applications aren't built to expect `keventd` :sweat_smile: so do we need a `kevent settle` command, or...? any suggestions, thoughts, etc... are greatly appreciated, as well as your experiences so far :bowing_man:

troglobit:
Three good points!

libudev-zero: you are right, I was mislead by "a common friend" :sweat_smile: I'll fix the docs. The Gardenhouse fork that adds `/run/udev/data/` support looks interesting; I hope it lands upstream.

Zombies from missing helpers: bug. I'll make sure to refactor to handle this better, thanks for taking the time to test and report!

Regarding settle: I think we have a really good abstraction with conditions in Finit, so my idea has been to to declare device requirements in the service file. For example:

    service <dev/dri/card0> /usr/bin/greetd -- Login manager

Good point about settle being discouraged, I had missed that and just tried to build something that wouldn't require it. Honestly, I didn't quite understand how it worked anyway.  Investigating it a bit more it seems they've deprecated it because it races and slows (serializes) the boot process.

Unknowingly I went my own merry way that turned out to be right.  E.g., I've added `<dev/eth0>` support, in addition to the `<net/eth0>` support we already have. The former fires on interface existing while the latter on link up.  I plan to add two more condition classes for hardware without a `/dev` node, like a switch core or sensor, but they'll follow sort of the udev way:

  - `class/<subsystem>/<name>`: fires when a sysfs class device appears
  - `bind/<driver>`: fires when a driver binds to its device

For your greetd case, `<dev/dri/card0>` should already work once you wrap it as a Finit service (minus bugs).

One more thing about keventd's design: it folds the common udev helpers (`path_id`, `usb_id`, `blkid`, `input_id`, `net_id`, `hwdb`, `kmod`) into the daemon as compiled-in builtins.  Less fork/exec at boot, single binary to audit.  Rules that call /lib/udev/<helper>; first check the path, and if missing fall back to the matching builtin.  So you can drop your own `/lib/udev/foo` to override a builtin, and missing helpers no-op (after the fix above) instead of failing loud.  Since I have no fido device to test with, so I'll leave `fido_id` as an exercise to the reader.

aanderse:
> Regarding settle: I think we have a really good abstraction with conditions in Finit, so my idea has been to to declare device requirements in the service file.

i think you have a very concrete way to solve my problem, i understood this and maybe should have elaborated... what happens if my graphics device flip flops between `card0` and `card1`? is there a more general way to depend on something coming up? i need to do more research into how udev handles this... 🤔


> plan to add two more condition classes for hardware without a `/dev` node

very cool!


> One more thing about keventd's design:

ahh! so this is a consistent design in keventd... wonderful! i had noticed it for device ownership and permission assignment, though not for these common helpers - very cool! i think i will go through the code again a bit deeper... all sorts of wonderful things exist in there 😃

thanks @troglobit!

troglobit:
> > Regarding settle: I think we have a really good abstraction with conditions in Finit, so my idea has been to to declare device requirements in the service file.
> 
> i think you have a very concrete way to solve my problem, i understood this and maybe should have elaborated... what happens if my graphics device flip flops between `card0` and `card1`? is there a more general way to depend on something coming up? i need to do more research into how udev handles this... 🤔

Yeah that sounds like you should match on a devpath or similar and create a symlink in an udev rule.  We do similar things for gps devices that can be plugged into any USB port, or the other way around provide usb-to-ttl converters with a fixed device name based on the USB port they are plugged into.

> > One more thing about keventd's design:
> ahh! so this is a consistent design in keventd... wonderful! i had noticed it for device ownership and permission assignment, though not for these common helpers - very cool! i think i will go through the code again a bit deeper... all sorts of wonderful things exist in there 😃
> thanks @troglobit!

Glad you like it! :relaxed:

aanderse:
still running `keventd`... it is great :+1: i played around with `libudev-zero` source a bit... got it to read tags from `/run/udev/data`... that was neat when paired with [pam_uaccess](https://git.sr.ht/~kennylevinsen/pam_uaccess)

one question: did you want `keventd` to do `pid` notification protocol, or none is fine because what does "ready" even mean for `keventd`?

troglobit:
> still running `keventd`... it is great 👍 i played around with `libudev-zero` source a bit... got it to read tags from `/run/udev/data`... that was neat when paired with [pam_uaccess](https://git.sr.ht/~kennylevinsen/pam_uaccess)

Wow, by now you've probably logged more miles than me on it :sweat_smile:  Any chance you can share the libudev-zero patch? I'd love to pull it into my Infix branch.

> one question: did you want `keventd` to do `pid` notification protocol, or none is fine because what does "ready" even mean for `keventd`?

Good question, the idea was to provide a reliable hand-over point "`/dev` is populated, persistent symlinks are live".  However, I realized that this had not yet been put into code, so the version you have created the pidfile after coldplug had triggered the kernel to re-emit events, but before keventd had drained them.  The changes I just pushed aim to fix that.

In Infix I lean on Finit conditions to set up "barriers".  Most services depend on `<pid/syslog>`, which can be provided by BusyBox syslogd + klogd or my sysklogd project, and syslogd itself depends on `<service/keventd/ready>`.  That chain serializes execution for brief periods at boot, which on Infix is partly there to make sure no logs leak to the console and get lost from log files. 

Separately, I've added a `keventd -S` one-shot command, similar to `udevadm settle` that just polls `/sys/kernel/uevent_seqnum` and exits when it's quiet.  Mostly for legacy boot scripts that can't be wired to conditions, but might be handy for your testing too.

aanderse:
> Any chance you can share the libudev-zero patch?

[sure!](https://github.com/aanderse/libudev-zero) i only added tag filtering from the udev database... i think if someone were to fill in more udev function stubs there is better ways to structure this code though... :sweat_smile:

did you implement any other udev functions?


---

i'll try the new version and report back - thanks!

troglobit:
> > Any chance you can share the libudev-zero patch?
> [sure!](https://github.com/aanderse/libudev-zero) i only added tag filtering from the udev database... i think if someone were to fill in more udev function stubs there is better ways to structure this code though... 😅

Thanks :+1: 

> did you implement any other udev functions?

Nothing more than the two new condition classes I mentioned before, 5a86ff8 covers that.

> i'll try the new version and report back - thanks!

Much appreciated!

aanderse:
still running... no problems, loving it :smile:

please please please merge this before `5.0` because this is the _perfect_ device manager for `finix`! :tada:

troglobit:
> still running... no problems, loving it 😄
> 
> please please please merge this before `5.0` because this is the _perfect_ device manager for `finix`! 🎉

Wow, that's high praise, thank you! :bowing_man: 

Actually, 5.0 might be closer than you think ... Swedish Midsummer is just around the corner, and with that my vacation.  Although a lot of time this year will go to painting and renovations (house), I will need some stimulating code sessions as well.  So hang in there! :sweat_smile:

aanderse:
i was trying to run my system by linking against the `gardendevd` fork of `libudev-zero` and realized that in `udevdb.c` `udevdb_devid()` keys net devices as `+net:<sysname>` - libudev expects `n<ifindex>`

this means net devices are never found in `/run/udev/data`

my theory is that if `keventd` follows what the other implementations do then it might be possible to run `networkmanager` on `keventd`!

troglobit:
> i was trying to run my system by linking against the `gardendevd` fork of `libudev-zero` and realized that in `udevdb.c` `udevdb_devid()` keys net devices as `+net:<sysname>` - libudev expects `n<ifindex>`
> 
> this means net devices are never found in `/run/udev/data`

Should be fixed now.  The branch is rebased on the new master with new .conf syntax.

> my theory is that if `keventd` follows what the other implementations do then it might be possible to run `networkmanager` on `keventd`!

Interesting! :thinking::bulb:

aanderse:
i found what i think are a few more inconsistencies between `/run/udev` with `keventd` vs `udev`... i will try to provide some additional details at some point

what is your plan for merging this branch? considering that nothing _currently_ uses `/run/udev` i think you can safely merge this as-is and state that `/run/udev` data is currently "experimental"... especially since nothing will make use of it yet

does that seem reasonable?

troglobit:
> i found what i think are a few more inconsistencies between `/run/udev` with `keventd` vs `udev`... i will try to provide some additional details at some point

That would be great, thank you!

> what is your plan for merging this branch? considering that nothing _currently_ uses `/run/udev` i think you can safely merge this as-is and state that `/run/udev` data is currently "experimental"... especially since nothing will make use of it yet
> 
> does that seem reasonable?

Yeah it does.  However, I'll rebase this branch on `master` as soon as I've finalized and merged the d-bus support I'm currently working on.  So it'll unfortunately take a bit more time.

troglobit:
Remaining work before merge:
 - Test on real¹ hardware
 - Verify D-Bus interface

____
¹) meaning a real monster

aanderse:
define monster? how many cores? RAM? do you have a machine to test on or should i ask around? 👀

troglobit:
> define monster? how many cores? RAM? do you have a machine to test on or should i ask around? 👀

It's a quad-core A72 with three cascade-connected DSA switches external PHY's with firmware etc. that we've had problems probing and loading before using eudev.  So it's a good canary :bird: to test on.

You're of course welcome to test it out yourself!  Just wanted to signal the end of the development work on this branch somehow :blush:

aanderse:
oh ok fantastic! so excited to see this merged! thanks

troglobit:
Status after testing on the monster: :heavy_check_mark:

PR #494, opened 2026-07-31
doc: Update link in CONTRIBUTING.md

henziger:
I noticed that the Pro Git link had gone stale, but found the new url for the same section at https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines. Not sure how stable that anchor is, time will tell I guess, but for now it seems take me as close as possible to the relevant section in the book.

NB: The link to the Kernel Normal Form Wikipedia page is stale as well, but I couldn't find any good replacement for it so I left it as is. :slightly_smiling_face:


troglobit (approved):
Awesome, thank you so much! 🙇‍♂️

PR #498, opened 2026-08-12
Misc cgroup fixes

### LINKED ISSUE #497: [feature request] cgroup delegation insufficient for lxd/incus?
i tried to use `delegate = true` with `incus` and noticed that it doesn't quite seem to work... `incus` fails to apply things like `cpuset`, for example, warning/errors like `Unable to set cpuset ... value="0,1,2,3,4,5"`

`finit` is doing what it set out to do, but i think `incus` has certain expectations - i _think_ the problem is that `finit` needs to create a private cgroup namespace because `incus` is trying to reparent to the cgroup of pid 1 🤔

i'm not an expert at `incus` internals, _and_ i didn't dig through the `finit` code on delegation... so there is a bit of hand waving on my side here... does this make sense though?

troglobit:
Possibly fixes #497

### DIFF
--- a/doc/config/cgroups.md
+++ b/doc/config/cgroups.md
@@ -42,7 +42,9 @@ the above three.  We leave `init/` and `user/` as-is reducing weight of
     cgroup maint  { cpu.weight = 100  }
 
 By default, the `system/` cgroup is selected for almost everything.  The
-`init/` cgroup is reserved for PID 1 itself and its closest relatives.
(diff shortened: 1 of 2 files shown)


aanderse:
can confirm that `incus` works great on my system with this :rocket: thank you @troglobit!

troglobit:
> can confirm that `incus` works great on my system with this 🚀 thank you @troglobit!

Excellent! Then that's the way to go, thank you for taking the time to verify! <3

Issue #420, opened 2024-12-09
Could a finit service run with a specific PAM config?


liuming50:
Our company is considering involve in PAM (https://en.wikipedia.org/wiki/Linux_PAM) in our system and we are using finit as init manager.

Could it be possible we run a service/task under a specific PAM config? For instance, in systemd, it has a PAMName support, if we set it in weston.service:

```
PAMName=weston-autologin
```

it will look for /etc/pam.d/weston-autologin:

```
auth      required  pam_nologin.so
auth      required  pam_unix.so     try_first_pass nullok

account   required  pam_nologin.so
account   required  pam_unix.so

session   required  pam_env.so
session   required  pam_unix.so
-session  optional  pam_systemd.so type=wayland class=user desktop=weston
-session  optional  pam_loginuid.so
```

and set the permission controls for weston service.

Would finit like to support this or is there a plan for supporting it?

troglobit:
First, there is nothing planned wrt. this.

Second, even though I just skimmed through the systemd docs on the topic, how do you imagine this all working?  In systemd they fork off a `(sd-pam)` service for each unit that has this directive, and it seems "the main unit process will be migrated to its own session scope unit" -- I don't understand what that means?  It's confusing to read and just seems like a lot of added complexity.  What is the benefit of having Finit support when the process can call [pam_start(3)](https://linux.die.net/man/3/pam_start) with the `weston-autologin` service?

troglobit:
I had a chat recently with a colleague who got intrigued by this idea.  Even though I still cannot see it, he usually has his wits about him, and I'm curious to understand how this could be added to Finit and what it could improve for users.  So please explain a bit more if you can.

liuming50:
@troglobit 

Thanks for the reply.

What I was asking basically is this feature in systemd: https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#PAMName=

wkz:
(I think I am the colleague that, apparently at least _usually_, has his wits about him :wink:)

We have read through the systemd docs. I think what we were curious about was _when_ do you typically need this? I.e., what are some typical use-cases?

liuming50:
> (I think I am the colleague that, apparently at least _usually_, has his wits about him 😉)
> 
> We have read through the systemd docs. I think what we were curious about was _when_ do you typically need this? I.e., what are some typical use-cases?

We have a system that have multiple users, and for a specific user, we want it has a specific permission setting.

liuming50:
[...]

Issue #301, opened 2022-08-23
Add support for running not only as PID 1


troglobit:
The process monitoring capabilities of Finit could be beneficial in use-cases other than just PID 1.  One such example use-case is http://supervisord.org/

aanderse:
i would get great use out of this feature off it existed

realistically is this on your roadmap at all?

troglobit:
> realistically is this on your roadmap at all?

It is, not short-term though.  There's quite a few things I'd like to refactor and add support for, and since much of it will break backwards compatibility, it has been added to the [5.0 milestone](https://github.com/troglobit/finit/milestone/11).

aanderse:
thanks for the information @troglobit

i will admit i have become [quite interested](https://github.com/aanderse/finix) in this excellent project of yours and it has filled a good chunk of my hobby time

purely for interest sake i would love if you could share any realistic development plans - to me it looks like the milestone issues you have cover other people's desires, not just yours... and since this appears to be at least partially a hobby for you too i don't expect those milestone issues too be something you work on

then again, maybe you just scratch itches as they come along and you don't have any hard plans or personal deadlines... in which case i will just enjoy new releases and features as they come 😄

thanks again!

aanderse:
hey @troglobit i hate to _ping_ on something like this but given how responsive you have been to all of my many other inquiries i thought it entirely possible that you simply missed my message/question above about your realistic development plans for `finit`

if you have any insight to share i appreciate, and if not then of course no worries - thanks!

troglobit:
Cool project there, Finix, I like it! 💪 💯 

Heh, I *had* missed this one! 😅 Sorry, but I get so many notifications ... now, regarding your inquiry.  The only concrete plans that exist are the milestones.  I have some reminders and notes to myself here and there, but nothing concrete.  Finit was sprung out of a need at a previous company I worked at, when a friend and I left we created [Infix](https://github.com/kernelkit/infix), with Finit as its PID 1.  So that's what mostly drives Finit development today -- requirements in Infix, customer requests, and when the team finds something in the system that could be done in a better way.

Sure, from time to time, users submit ideas and half-baked implementations that can spark sudden bursts of creativity and open up new opportunities that need exploring.  I really want it all to fit together nicely and feel right to users and myself.

aanderse:
[...]

PR #500, opened 2026-08-20
Add support for running services inside a PAM session

troglobit:
Apply a PAM session to run/task/sysv/services Finit starts, pam_limits above all, so a service running as a given user picks up that user's limits the way a login does.

Add a new `pam` setting for the new block format (only), like the per-service directories, naming a file in /etc/pam.d:

    service weston {
        user    = "weston"
        pam     = "weston-autologin"
        command = "/usr/bin/weston --continue-without-input"
    }

`pam_close_session()` has to be called by a process still holding the handle, and the handle does not survive `exec()`.  Hence the keeper: it holds the handle, drops to the service's credentials, and waits for a parent-death signal before closing the session.  Same shape as systemd's (sd-pam), for the same reason, and one per fork, so the script hooks open and close their own.

The keeper closes the descriptors it inherited from Finit and only those.  Closing everything would also take out what `pam_open_session()` opened for itself, a keyring fd or a lock file, and leave the modules to close a session with those pulled out from under them.  Closing nothing, as (sd-pam) does, would leave it holding the write end of the notify pipe for the service's whole lifetime and starve `notify = "s6"` services of their ready signal.  So the fds open before `pam_start()` are snapshotted and exactly those are closed, while the ones PAM opens after are marked close-on-exec so the daemon does not inherit them either.

A refused value, a denied account stack, an uninstalled `pam.d/` file, and a build without PAM support all keep the service from starting rather than running it with the stacks skipped: one that quietly loses `pam_limits` and its private `/tmp`, with nothing said.  Capabilities a module like `pam_cap.so` granted are merged into the IAB Finit applies instead of being replaced by it, which only helps a service that also sets capabilities, the other arm being a plain `setuid()` with nothing left to restore once permitted is empty.

The test sysroot gains `pam_permit.so`, `pam_deny.so` and `pam_limits.so`, which ldd cannot see, libpam `dlopen()`s them, and the test skips when the host has none to stage.  The negative cases pin the exit status rather than only asserting crashed, which `serv` reports for any early exit, so a bad command or an unwritable pidfile cannot pass for a rejected session.


troglobit:
@liuming50 here's a proposal for your feature request #420 slated for v5.0.  I hope it meets your expectations, and I'd be very grateful if you could have a look and let me know what you (all) think about the shape and implementation.

troglobit:
Ping @liuming50 ... hope you see this. We're adding support for user sessions/services #301,  and this is one critical piece of the puzzle, so I'd really appreciate your feedback before I merge this! Thank you in advance :pray:

liuming50:
@troglobit thanks for the great work! I would review it and have a try as well. Will get back to you soon.

troglobit:
> @troglobit thanks for the great work! I would review it and have a try as well. Will get back to you soon.

Awesome, looking forward to your feedback! 😃

liuming50:
Hi, @troglobit 

Sorry for the delay, I was busy on some other things, I have finally done some tests as below:

Check if login user running in PAM:

```
$ grep -i pam /var/log/auth.log
2026-09-21T08:43:29.646590+00:00 (none) login[1159]: pam_unix(login:session): session opened for user tester(uid=1000) by LOGIN(uid=0)
```

Check if services (started by finit) are running with libpam:

```
$ pidof crond
707
$ sudo grep -i pam /proc/707/maps
754e6000-754ef000 r-xp 00000000 fe:00 5457       /lib/libpam.so.0.85.1
754ef000-754f0000 r--p 00008000 fe:00 5457       /lib/libpam.so.0.85.1
754f0000-754f1000 rw-p 00009000 fe:00 5457       /lib/libpam.so.0.85.1
```

$ pidof sshd
425
$ grep -i pam /proc/425/maps
76ef1000-76efa000 r-xp 00000000 b3:02 1131       /lib/libpam.so.0.85.1
76efa000-76efb000 r--p 00008000 b3:02 1131       /lib/libpam.so.0.85.1
76efb000-76efc000 rw-p 00009000 b3:02 1131       /lib/libpam.so.0.85.1


Check if a service can run with a PAM config:

finit config:

```
service connmand {
    pam = "connmand"
    command = "/usr/sbin/connmand -n"
}
```

see logs like:

```
/var/log/auth.log:2026-09-22T10:52:34.711753+00:00 (none) finit[466]: pam_unix(connmand:session): session opened for user root(uid=0) by (uid=0)
```

so seems is working fine?

troglobit:
> Hi, @troglobit

Hiya! :smiley: 

> Sorry for the delay, I was busy on some other things, I have finally done some tests as below:

No worries at all, been busy myself.

> [snip]
> so seems is working fine?

Very happy to hear this, thank you for taking the time to verify! :bowing_man:

PR #501, opened 2026-08-21
libsystemd: guard sd-daemon.h with extern C for c++ consumers

aanderse:
i tried to build `smartmontools` with `finit` instead of `libsystemd` because all it uses is `sd_notify`... and found this!


troglobit (approved):
Awesome! 🙇‍♂️

